PR #1362, opened 2014-12-11
Swallow EPIPE errors

weswigham:
When thrown in a named pipe input channel - usually on pipe connection end. 

Under the current scheme, functions like `RoleEnvironment.getConfigurationSettings` take a callback which handles an error or a success. Except there's a problem - assuming you do handle the error (and don't kill the process because of it) the handler will get called TWICE. And the second time will be with the actual configuration data (as the first was passed the EPIPE shutdown that was thrown). 

While, for a lot of architectures, this simply means the EPIPE error gets logged or otherwise dealt with, then data handled, this is unacceptable when wrapping the API in promises, ala `Promise.denodify`. If I'm not mistaken, many of these callbacks are meant to be triggered once and considered done - not reused.

This might not be the way to change this to support this, but considering a `shutdown EPIPE` is always thrown on connection close (which `NamedPipeInputChannel` does automatically), it's probably not the worst idea.

Though, this may also be a good opportunity to ask how anyone feels about updating the API to work with promises correctly, rather than needing to wrap it.

### DIFF
--- a/lib/serviceruntime/namedpipeinputchannel.js
+++ b/lib/serviceruntime/namedpipeinputchannel.js
@@ -76,6 +76,7 @@ NamedPipeInputChannel.prototype._readData = function (name, callback) {
   });
 
   self.client.on('error', function (error) {
+    if (error.code == 'EPIPE') { return; } //Swallow EPIPE errors emitted when the connection is closed
     callback(error);
   });
 };


lib/serviceruntime/namedpipeinputchannel.js
```diff
@@ -76,8 +76,9 @@ NamedPipeInputChannel.prototype._readData = function (name, callback) {
   });
 
   self.client.on('error', function (error) {
+    if (error.code == 'EPIPE') { return; } //Swallow EPIPE errors emitted when the connection is closed
```
yugangw-msft: ===

yugangw-msft:
could you please provide some background such as an end to end case to describe how your change will make the difference, say you will be able to catch the error your care and safely ignore EPIPE?

weswigham:
This is from memory, so take it with a bit of a grain of salt. I believe at the time I was using it to communicate with a local azure storage emulator, and I had wrapped the API up in promises using `Promise.denodify` from the npm `promise` library. I found that every short-lived connection to the storage emulator when the connection ended threw an EPIPE to signal the end of connection, causing promises which should have passed (the data had already been received internally) to fail, due to the connection closing before the data return callback got called - ie the callback would get called twice, one with an EPIPE and once with the data, in that order. Promises are single-use-only, so this broke my wrapper pretty handily. I believed simply swallowing it at this step should have been harmless, HOWEVER, looking at it now, I think I can propose a better change:

In this file the data is handled like so:
```
 self.client.on('data', function (data) {
     if (self.closeOnRead) {
       self.client.end();
     }
 
     callback(undefined, data);
   });
```
Reordering it to this would fix the issue from the promise-consumption perspective, as the data would be consumed before the error:
```
 self.client.on('data', function (data) {
     callback(undefined, data);

     if (self.closeOnRead) {
       self.client.end();
     }
   });
```
Though this may have other implications, it may be safer than simply swallowing EPIPE errors when the connection is closing - though there probably shouldn't be any harm in doing that, either.

yugangw-msft:
we have ported your initial change to the master branch and published a new common package. So close this one. 
Thanks for the contributions.

PR #1367, opened 2015-01-21
Specify pem file encoding

anthonye2007:
Without the encoding specified, the pem file is read in as a buffer instead of a string.
This would then fail the argument validation requirement for credentials.pem since it wanted a string and got a buffer.
Without the encoding I was unable to run this example.
With the encoding specified, I am able to run the example.


azurecla:
Hi __@anthonye2007__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
Thanks for the contribution.

PR #1372, opened 2015-01-26
updated package versions in packages.config to generate code from latest specs

amarzavery:


amarzavery:
@yugangw-msft - Should we also update the versions in package.json?

Issue #1416, opened 2015-05-07
XML namespace error in dac.importMethod


vincent:
There is an XML namespace error in the [importMethod](https://github.com/Azure/azure-sdk-for-node/blob/14a19d14f1ca76766bb24d206d1f19471f9f7f16/lib/services/sqlManagement/lib/sqlManagementClient.js#L214) code

Currently the request fails with error 

```
Error: The Uri doesn't match any specific service operation.
```

When I hard-code the XML request content, taken from https://msdn.microsoft.com/en-us/library/azure/dn781284.aspx, that is with the namespaces, it works correctly.

current produced XML:

``` xml
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<ImportInput xmlns="http://schemas.datacontract.org/2004/07/Microsoft.SqlServer.Management.Dac.ServiceTypes">
    <AzureEdition>Business</AzureEdition>
    <BlobCredentials type="BlobStorageAccessKeyCredentials">
        <Uri>the_blob_uri</Uri>
        <StorageAccessKey>the_blob_acccesskey</StorageAccessKey>
    </BlobCredentials>
    <ConnectionInfo>
        <DatabaseName>the_db_name</DatabaseName>
        <Password>the_db_password</Password>
        <ServerName>the_db_server</ServerName>
        <UserName>the_db_user@the_db_server</UserName>
    </ConnectionInfo>
    <DatabaseSizeInGB>10</DatabaseSizeInGB>
</ImportInput>
```

hard-coded XML template:

``` xml
<ImportInput xmlns="http://schemas.datacontract.org/2004/07/Microsoft.SqlServer.Management.Dac.ServiceTypes" xmlns:i="http://www.w3.org/2001/XMLSchema-instance">
  <AzureEdition>Web</AzureEdition>
  <BlobCredentials i:type="BlobStorageAccessKeyCredentials">
    <Uri>{uriBase}</Uri>
    <StorageAccessKey>+Ka1LjBc36Pa62WOCJ9zXJjlN5qU6YU3B2JYTgSfw8xiQKlrBoOmAEcdJSO3/YPA5JcJAmMSWCyQB7w1D96k8A==</StorageAccessKey>
  </BlobCredentials>
  <ConnectionInfo>
    <DatabaseName>{dbName}</DatabaseName>
    <Password>{myPassword}</Password>
    <ServerName>{serverName}.database.windows.net</ServerName>
    <UserName>{mylogin}</UserName>
[...]

PR #1418, opened 2015-05-07
add a namespace attribute to the root node for #1416

vincent:
A missing namespace in the XML request makes the `sqlmgmt.dac.importMethod()` to fail. see #1416

This PR naively adds the namespace as an attribute of the root node.


azurecla:
Hi __@vincent__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
@akromm, please take a look and see whether we should address at the spec level

vincent:
I understand this file is generated, and that this change may not be clean or even wanted. I just completed the diff to make it usable (technically)

yugangw-msft:
I opened an issue #1432 to track this and give it to adam to evaluate.

PR #1419, opened 2015-05-11
Update ServiceBus Readme to include an example

devigned:
- Add example from [node-cerulean](https://github.com/noodlefrenzy/node-cerulean/blob/master/lib/index.js) to help users get started.

### DIFF
--- a/lib/services/serviceBus/README.md
+++ b/lib/services/serviceBus/README.md
@@ -14,9 +14,66 @@ npm install azure-sb
 
 ## How to Use
 
-### Authentication
+```node
+var azure = require('azure');
 
-### Create the ServiceBus service
+function checkForMessages(sbService, queueName, callback) {
+    sbService.receiveQueueMessage(queueName, { isPeekLock: true }, function (err, lockedMessage) {
+        if (err) {
+            if (err == 'No messages to receive') {
+                console.log('No messages');
+            } else {
+                callback(err);
+            }
+        } else {
+            callback(null, lockedMessage);
+        }
+    });
+}
+
+function processMessage(sbService, err, lockedMsg) {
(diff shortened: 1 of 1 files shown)


devigned:
@noodlefrenzy, @djrosanova showed this to me and I thought it would be awesome to include. I've attributed you as the original source, but if you prefer that it not be included here, we'll not merge the PR.

noodlefrenzy:
@devigned I love it!  Include away :)  Especially since eventually I'll be moving/removing it since node-cerulean is destined for other (nefarious) purposes. (TMI version: Specifically initially I need to wrap blob leases to make them more node-like and let me build something like EventProcessorHost into node-sbus.)

devigned:
@noodlefrenzy thank you for the awesome work! After node-cerulean goes away, I'll drop the link, but leave the attribution.

noodlefrenzy:
@devigned Glad to help! I'll let you know when I make the change, and just move it to an examples directory.  Also, I should probably let @djrosanova know that I accidentally left it running for 3 days and never saw any issues :)

lib/services/serviceBus/README.md
```diff
@@ -14,9 +14,66 @@ npm install azure-sb
 
 ## How to Use
 
-### Authentication
+```node
+var azure = require('azure');
 
-### Create the ServiceBus service
+function checkForMessages(sbService, queueName, callback) {
+    sbService.receiveQueueMessage(queueName, { isPeekLock: true }, function (err, lockedMessage) {
```
yugangw-msft: if you still want it to be merged, please get the indent to be 2 spaces

yugangw-msft:
close, and please fix the indent and submit to master when it is ready

PR #1421, opened 2015-05-22
pump up the version so to test publish through jenkins ci

yugangw-msft:
This is a temporary package owned by us, so use it to test new ci job to publish to npm


yugangw-msft:
the CI failure is a flakiness, which has nothing to do this change. Will forward to CSM team
"arm group template show should show a resource group template from gallery with positional name:
     Uncaught Error: Nock: No match for HTTP request GET /Microsoft.Gallery/galleryitems/Microsoft.WebSiteMySQLDatabase.0.3.15-preview "

PR #1434, opened 2015-06-09
Update ServiceBus Readme to include an example

devigned:


lib/services/serviceBus/README.md
```diff
@@ -14,9 +14,66 @@ npm install azure-sb
 
 ## How to Use
 
-### Authentication
+```node
+var azure = require('azure');
 
-### Create the ServiceBus service
+function checkForMessages(sbService, queueName, callback) {
+    sbService.receiveQueueMessage(queueName, { isPeekLock: true }, function (err, lockedMessage) {
```
yugangw-msft: please make indent at 2

lib/services/serviceBus/README.md
```diff
@@ -14,9 +14,66 @@ npm install azure-sb
 
 ## How to Use
 
-### Authentication
+```node
+var azure = require('azure');
```
yugangw-msft: it is unclear this code has nothing to do with the Authentication, should we remove it?

lib/services/serviceBus/README.md
```diff
@@ -14,9 +14,66 @@ npm install azure-sb
...
+       if (err) {
+           console.log('Failed Tx: ', err);
+       } else {
+           console.log('Sent ' + msg);
+       }
+    });
+}
+
+var connStr = process.argv[2] || process.env.CONNECTION_STRING;
+if (!connStr) throw new Exception('Must provide connection string');
```
yugangw-msft: Use "Error"?

lib/services/serviceBus/README.md
devigned: the authentication line is deleted already

yugangw-msft: ah, didn't read it carefully.

PR #1435, opened 2015-06-10
Fixed case of main module entry point

christav:


azurecla:
Hi __@christav__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (ctavares). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/services/apiAppManagement/package.json
```diff
@@ -17,7 +17,7 @@
     "management",
     "apiapp"
   ],
-  "main": "./lib/apiapp.js",
+  "main": "./lib/apiApp.js",
```
yugangw-msft: yesterday, we just renamed the physical file name in the linux box to see it worked, but updating the entry in the  package.json is also good. merging

christav: Yeah, it turns out to be really hard to rename a file with just a case difference on a windows box. :-)

PR #1437, opened 2015-06-10
Updated setup.js script to automatically find services

christav:
Iterate over the services directory and read out the package names instead of maintaining that giant array.

Also adds a teardown script to remove the global changes done by setup.js.

### DIFF
--- a/scripts/each-service.js
+++ b/scripts/each-service.js
@@ -1,0 +1,137 @@
+/**
+* Copyright (c) Microsoft.  All rights reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+'use strict';
+
+// Helper script that can read the set of services included
+// in the azure SDK dynamically.
+
+var fs = require('fs');
+var path = require('path');
+var util = require('util');
+
+var sdkRoot = path.join(__dirname, '..');
+var servicesRoot = path.join(sdkRoot, 'lib/services');
+
+//
+// Can't use readable-stream, have to only use core modules
+// since this is called by ci before npm install runs.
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@christav__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (ctavares). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

scripts/each-service.js
```diff
@@ -0,0 +1,118 @@
...
+  fs.readdir(servicesRoot, function (err, files) {
+    if (err) { return done(err); }
+    done(null, [path.join(sdkRoot, 'lib/common')].concat(
+      files.map(function (f) {
+        return path.join(servicesRoot, f);
+      })));
+  });
+}
+
+function hasPackageJson(servicePath, done) {
```
yugangw-msft: This routine appears just to check the service path exists, rather check the package.json exists. I think we should add code to check the file exists. Checking the folder might not be meaningful as the collection were returned from  "fs.readdir".

scripts/each-service.js
```diff
@@ -0,0 +1,118 @@
...
+      }, index, items);
+  }
+  doReduce(initialValue, items, 0);
+}
+
+function filterAsync(items, asyncPredicate, done) {
+  reduceAsync(items,
+    function(acc, item, next) {
+      asyncPredicate(item, function(err, passed) {
+          if (err) { return next(err); }
```
yugangw-msft: indent?

scripts/each-service.js
```diff
@@ -0,0 +1,118 @@
+// Helper script that can read the set of services included
```
yugangw-msft: nit, license header?

christav: @yugangw-msft None of the other scripts in this folder have a license header.

scripts/each-service.js
christav: I actually did this extra level of indentation deliberately in this case. I was getting lost in all the nested callbacks. Basically, the code is indented as if the preceding "function() {" was on it's own line. I found it much easier to read this way. I can bump the functions down a line to make everything line up if you'd like.

scripts/each-service.js
christav: Good catch. It's supposed to check for the package.json file. Will update.

christav:
@yugangw-msft I've updated according to your comments, please re-review.

PR #1441, opened 2015-06-17
Azure Key Vault client, version 0.10.

msfcolombo:
This change introduces a new version of the Key Vault client for NodeJS.


azurecla:
Hi __@msfcolombo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (fcolombo). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

lib/azure.js
```diff
@@ -645,6 +645,8 @@ exports.createMetricsClient = azureMonitoring.createMetricsClient;
  */
 var azureKeyVault = require('azure-keyvault');
 
+exports.KeyVault = azureKeyVault;
+
```
yugangw-msft: you can revert the change in this file, we are no longer maintain this roll up package

lib/services/keyVault/lib/keyVault.js
```diff
@@ -16,366 +16,677 @@
...
-* @returns {Stream} The response stream.
-*/
-KeyVaultSecrets.prototype.get = function (secretIdentifier, callback) {
-  return this.secrets.get(secretIdentifier, secretResponseDeserializer(callback));
+ * Performs a SET SECRET operation, which stores a secret in the service.
+ *
+ * @param {string}                              vault                 The vault identifier.
+ * @param {string}                              secretName            The secret name. If a secret with this name already exists, the service creates a new version and set as current.
+ * @param {string}                              request.value         The secret value.
+ * @param {string}                              [request.contentType] The value's content type.
```
yugangw-msft: this doesn't match the actual method signature

lib/services/keyVault/lib/keyVault.js
```diff
@@ -16,366 +16,677 @@
...
-* @param {Number} [top]
-* @param {function} callback
-* @returns {Stream} The response stream.
-*/
-KeyVaultSecrets.prototype.list = function (vault, top, callback) {
-  return this.secrets.list(vault, top, secretResponseDeserializer(callback));
+ * Performs an UPDATE SECRET operation, which updates attributes of a secret version.
+ *
+ * @param {string}                              secretIdentifier      The secret identifier. It may or may not contain a version path. If a version is not provided, the latest secret version is used.
+ * @param {stri}                                [request.contentType] If informed, changes the value's content type.
```
yugangw-msft: this doesn't match the signature as well

lib/services/keyVault/lib/keyVault.js
```diff
@@ -16,366 +16,677 @@
...
+    result = JSON.parse(result.rawJsonRequest);
+    attrDatesToNumbers(request.attributes, result.attributes);
+    result = { rawJsonRequest: JSON.stringify(result) };
+  }
+  return result;
+}
+
+function valueResponseDeserializer(callback) {
+  return keyResponseDeserializer(function(err, result, response) {
+    if (err || !result || !result.value) {
```
yugangw-msft: this is a little questionable to me. are you return on a error case? If yes, then it is possible the "err" is null, and the downstream code might take a a success case. Otherwise, this code is good.

test/services/keyVault/keyVault-key-tests.js
```diff
@@ -0,0 +1,658 @@
...
+        client.deleteKey(keyId.vault, keyId.name, function(err, keyBundle) {
+          if (err) throw err;
+          compareObjects(createdBundle, keyBundle);
+          next();
+        });
+      }
+
+      function getKeyReturnsNotFound(next) {
+        client.getKey(keyId.baseIdentifier, function(err, keyBundle) {
+          if (!err || !err.code || err.code !== 'KeyNotFound' || !err.statusCode || err.statusCode !== 404) throw new Error('Unexpected error object: ' + JSON.stringify(err, null, ' '));
```
yugangw-msft: please break this line. too long

test/services/keyVault/keyVault-key-tests.js
```diff
@@ -0,0 +1,658 @@
...
+      series([
+        importToSoftware,
+        importToHardware,
+        function() {done();}
+      ]);     
+
+    });
+  });
+ 
+  xdescribe('list', function() {
```
yugangw-msft: is this the right suite ? 'x'?

test/services/keyVault/keyVault-secret-tests.js
```diff
@@ -0,0 +1,377 @@
...
+  
+  describe('CRUD operations', function () {
+    it('should work', function (done) {
+      
+      var createdBundle;
+      var secretId;
+
+      function createSecret(next) {
+        client.setSecret(vaultUri, SECRET_NAME, { value: SECRET_VALUE }, function(err, secretBundle) {
+          if (err) throw err;
```
yugangw-msft: do you have to throw, or just give the error to callback which is recommended, like "return next(err)". I believe the "next" should be the 'callback', right?

lib/azure.js
msfcolombo: ok

test/services/keyVault/keyVault-secret-tests.js
```diff
@@ -0,0 +1,377 @@
...
+
+        deleteASecret();
+      }
+
+      series([
+        createManySecrets,
+        listSecrets,
+        deleteSecrets,
+        function() { 
+          if (!suiteUtil.isMocked) {
```
yugangw-msft: i think you should use "isPlayback"

lib/services/keyVault/lib/keyVault.js
msfcolombo: Good catch. I renamed the parameter on JsDoc, but forgot to change the implementation. Fixing.

lib/services/keyVault/lib/keyVault.js
msfcolombo: Same as above.

test/services/keyVault/kv-test-utils.js
```diff
@@ -0,0 +1,267 @@
...
+  while (leadingZeroes < data.length && data[leadingZeroes] === 0) {
+    ++leadingZeroes;
+  }
+  if (leadingZeroes) {
+    data = data.slice(leadingZeroes);
+  }
+  return new Buffer(data);
+}
+
+function xdescribe(a,b,c) {
```
yugangw-msft: have you tried "describe.skip"?

lib/services/keyVault/lib/keyVault.js
msfcolombo: This is just a callback wrapper. We want to forward all errors to the inner callback.

test/services/keyVault/keyVault-key-tests.js
msfcolombo: sure

test/services/keyVault/keyVault-key-tests.js
msfcolombo: Yes, the test is disabled. I will add a comment.

test/services/keyVault/keyVault-secret-tests.js
msfcolombo: I prefer throwing. This is the fail fast approach. The stacktrace becomes more helpful.

test/services/keyVault/kv-test-utils.js
msfcolombo: Nobody else uses skip. I tried it and it polutes output with "pending", so can I leave the xdescribe?

test/services/keyVault/keyVault-secret-tests.js
msfcolombo: I got undefined symbol. Where is it defined? Also nobody else seem to be using it. Perhaps my branch is outdated?

yugangw-msft: This is fine. I realize we are at sdk-for-node; not xplat.

test/services/keyVault/kv-test-utils.js
yugangw-msft: Please use the .skip as that is the mocha convention. Showing "pending" is nothing wrong; rather tell us there are tests being skipped. Similar with .net xunit that you get warnings on this.
Also please add comments for why we are skip them.

test/services/keyVault/keyVault-key-tests.js
yugangw-msft: as i mentioned later, please use ".skip". This coed will coexist with many other codes, it is important to be consistent

msfcolombo: done!

test/services/keyVault/kv-test-utils.js
msfcolombo: done!

package.json
```diff
@@ -78,7 +78,10 @@
     "grunt-jsdoc": "~0.5.1",
     "grunt-devserver": "~0.4.1",
     "azure-storage-legacy": "0.9.14",
-    "xmlbuilder": "0.4.3"
+    "xmlbuilder": "0.4.3",
+    "adal-node": "0.1.13",
```
yugangw-msft: could you please undo this?

msfcolombo: If I undo this, the key vault test code would break.

PR #1442, opened 2015-06-18
update azure-storage module to fix security issues

jbinard:
Update azure-storage to version 0.4.5 to fix the following security issues:
- https://github.com/Azure/azure-storage-node/issues/61
- https://github.com/Azure/azure-storage-node/pull/65

All tests seems good with this new version.

### DIFF
--- a/lib/common/package.json
+++ b/lib/common/package.json
@@ -37,7 +37,7 @@
     "underscore": "1.4.x",
     "tunnel": "~0.0.2",
     "request": "2.45.0",
-    "validator": "~3.1.0",
+    "validator": "~3.22.2",
     "envconf": "~0.0.4",
     "duplexer": "~0.1.1",
     "through": "~2.3.4"


azurecla:
Hi __@jbinard__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

yugangw-msft:
We don't plan to keep on publishing this roller-up package, hence I am closing this PR.
2 reasons:
1. To make it easier that each azure service team can publish their own package independently at any time, without coordinating and conflicting on the roll-up package versioning.
2. The huge amount of files brought in by this package makes the client takes a big perf hit which doesn't justify with the fact that most of them only uses a tiny portion.

So the client is encouraged to only pull the specific package it needs.

PR #1479, opened 2015-08-10
update package.json file of arm-storage package

yugangw-msft:


yugangw-msft:
CI failure was expected as the test infrastructures is still ongoing.

PR #1485, opened 2015-08-17
test infrastructure improvements

amarzavery:


test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+ * 
+ * @param {Array} env - (Optional) Array of environment variables required by the test
+ * Example:
+ * [
+ *   { requiresToken: true },
+ *   { name: 'AZURE_ARM_TEST_LOCATION', defaultValue: 'West US' },
+ *   { name: 'AZURE_AD_TEST_PASSWORD'},
+ * ];
+ */
+function ArmTestUtils(mochaSuiteObject, testPrefix, env) {
```
yugangw-msft: rename to "SuiteBase"

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+  this.mochaSuiteObject = mochaSuiteObject;
+  this.testPrefix = this.normalizeTestName(testPrefix);
+  this.mockServerClient;
+  this.currentTest = '';
+  this.scopeWritten;
```
yugangw-msft: this is not used anywhere else

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+  _setupResourceManagementClient: function () {
+   this.resourceManagement = new ResourceManagementClient(this.createUserCredentials(), this.subscriptionId);
+   if (this.isPlayback) {
+    this.resourceManagement.longRunningOperationRetryTimeoutInSeconds = 0;
+   }
+  },
+
+  createResourcegroup: function (groupName, location, callback) {
+    console.log('Creating Resource Group: ' + groupName);
+    //if (!this.isPlayback) {
```
yugangw-msft: i assume you re using the "stub" option? So the commented out check will eventually be gone?

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+    //}
+    //return callback(null);
+  },
+
+  /**
+   * Creates a new UserTokenCredentials object.
+   * 
+   * @returns {ms-rest-azure.UserTokenCredentials} The user token credentials object
+   */
+  createUserCredentials: function () {
```
yugangw-msft: I think this can be folded into the constructor.

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+  getRecordingsDirectory: function () {
+    return this.recordingsDirectory;
+  },
+  
+  /**
+   * Provides the recordings directory for the test suite
+   *
+   * @returns {string} The test recordings directory
+   */
+  getTockenCacheFilePath: function () {
```
yugangw-msft: I suggest not to create the help method unless it is really needed. This can be folded into the constructor as well.

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
...
+  },
+  
+  normalizeTestName: function (str) {
+    return str.replace(/[{}\[\]'";\(\)#@~`!%&\^\$\+=,\/\\?<>\|\*:]/ig, '').replace(/(\s+)/ig, '_');
+  },
+  
+  normalizeEnvironment: function (env) {
+    env = env.filter(function (e) {
+      if (e.requiresCert || e.requiresToken) {
+        this.requiresCert = e.requiresCert;
```
yugangw-msft: we can remove the cert thing as we do not plan to support it

test/testlist.txt
```diff
@@ -1,74 +1,74 @@
...
-services/serviceBus/mpnsservice-registrations-tests.js
-services/serviceBus/notificationhubs-tests.js
-services/serviceBus/wnsservice-tests.js
-services/serviceBus/wnsservice-registrations-tests.js
-services/serviceBus/wrapservice-tests.js
-services/resourceManagement/resourceManagementClient-tests.js
-services/resourceManagement/authorizationClient-tests.js
-services/resourceManagement/featureClient-tests.js
-services/resourceManagement/subscriptionClient-tests.js
+#common/credentials/cert-tests.js
```
yugangw-msft: i assume we will create a new list for autorest based tests?

test/framework/arm-test-utils.js
```diff
@@ -0,0 +1,527 @@
+// 
+// Copyright (c) Microsoft and contributors.  All rights reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
```
yugangw-msft: We should use mit license header

yugangw-msft:
The CI failure looks legit that we might use not available on 0.8

test/framework/arm-test-utils.js
amarzavery: removed

PR #1527, opened 2015-11-03
minor fix to variable positioning for node sdk

begoldsm:
addDataLakeStoreAccount had a variable on the wrong side of an if during
auto generation.


begoldsm:
Cancelling this pull request, as we are fixing this issue in the CLI instead.

PR #1550, opened 2015-12-03
Update readme and remove follow redirect

begoldsm:
we have three functions that should not follow redirects. Fixing those
so that they don't.
Also adding the completed README documentation from our PM and Doc team.


amarzavery:
@begoldsm - once you publish the package, the README.md will be displayed in the [npm gallery](npmjs.com) for your packages.

PR #1574, opened 2015-12-29
ARM Networking: generating from latest spec

romangromov:
InboundNATRule updated

@DeepakRajendranMsft @amarzavery

### DIFF
--- a/lib/services/networkManagement2/generate.cmd
+++ b/lib/services/networkManagement2/generate.cmd
@@ -4,7 +4,7 @@
 ::
 
 @echo off
-set autoRestVersion=0.14.0-Nightly20151223
+set autoRestVersion=0.14.0-Nightly20151228
 if  "%1" == "" (
     set specFile="https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-network/2015-06-15/swagger/network.json"
 ) else (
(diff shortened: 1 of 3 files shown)


azurecla:
Hi __@romangromov__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (v-rogrom). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

DeepakRajendranMsft:
Hi @romangromov, it looks good to me. @amarzavery wanted to use new versions of ms-rest (1.4.0) and ms-rest-azure (1.4.0). Could you also update this in package.json?

amarzavery:
@romangromov The package.json version [here](https://github.com/Azure/azure-sdk-for-node/blob/autorest/lib/services/networkManagement2/package.json#L29) is `^1.2.0` which means anything greater than 1.2.0. So you should be good. You will have to manually cleanup node_modules in your devbox and execute npm install again for it to install 1.4.0. Because if it has 1.3.0 then it wont do anything because it satisfies the condition of being greater than 1.2.0

PR #1594, opened 2016-01-19
Update storage version

emgerner-msft:


lib/services/storageManagement2/package.json
```diff
@@ -11,7 +11,7 @@
     "Wang, Yugang <yugangw@microsoft.com>",
     "Zavery, Amar <amzavery@microsoft.com>"
   ],
-  "version": "0.12.0",
+  "version": "0.12.0-alpha",
```
amarzavery: I think it should be "0.12.0-preview", as it is a common pattern in the node sdk. I have submitted another PR https://github.com/Azure/azure-sdk-for-node/pull/1595 as I had published new version of the client runtime and wanted to regenerate using a new version of the code generator.

amarzavery:
@emgerner-msft - I think these two branches in upstream named 'emgerner-msft-patch-1' and 'emgerner-msft-patch-2' were created by mistake. You can take a look at them [here](https://github.com/Azure/azure-sdk-for-node/branches). If so, then should I delete them?

emgerner-msft:
I deleted emgernermsft-patch-1 (that was an accident), but patch-2 is this change so probably best not to delete that until merged.

emgerner-msft:
Ah sorry, didn't realize until just now that you updated version in the other pull request. Preview is fine too, the examples I found npm referencing used alpha but it's not important to me. I'll close this. Please mail me when you've merged the other change.

PR #1630, opened 2016-02-16
Added to Azure Service Bus examples ...

christophersw:
to demonstrate process for getting the current length of an Azure Service Bus Queue. 

I needed a method to analyze a service bus queue's length independent of reading/writing to it. It took me a while to realize that the queue object contained the properties I was looking for (albeit a bit cryptically named properties). I figured an example here might help others reach this conclusion a little faster.

Turns out the queue object contains a number of very helpful properties - so I included a JSON dump of those too. This may not be perfect documentation, but it is a lot better than the nothing that was here before.

### DIFF
--- a/examples/service-bus.md
+++ b/examples/service-bus.md
@@ -13,7 +13,7 @@ npm install azure-sb
 ### The basics of producers and consumers
 
 ```node
-var azure = require('azure');
+var azure = require('azure-sb');
 
 function checkForMessages(sbService, queueName, callback) {
   sbService.receiveQueueMessage(queueName, { isPeekLock: true }, function (err, lockedMessage) {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@christophersw__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla.azure.com.
    </span>

TTYL, AZPRBOT;

azurecla:
__@christophersw__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

PR #1637, opened 2016-02-26
Graph SDK generated from swagger specs

shuagarw:
- Graph client generated from swagger specs using graph version 1.6-internal


lib/services/graphManagement/README.md
```diff
@@ -0,0 +1,37 @@
...
+ var credentials = new msRestAzure.ApplicationTokenCredentials('your-client-id', 'your-domain', 'your-secret');
+ ```
+
+### Create the Graph Rbac client
+
+```javascript
+var fs = require("fs"),
+graphRbacManagement = require("azure-graph");
+
+var graphRbacManagementClient = graphRbacManagement.createGraphRbacManagementClient(new common.TokenCloudCredentials({
```
amarzavery: this is wrong.
it should be like this:
```javascript
var graphRbacManagementClient = require("azure-graph"); 
var graphClient = new graphRbacManagementClient(credentials, 'your-tenant-id');
```

PR #1638, opened 2016-02-29
AzureCP PR S58~S59 - Update Node SDK

huangpf:


amarzavery:
@huangpf  - Will this sample script work with the latest changes - https://github.com/Azure/azure-sdk-for-node/blob/autorest/examples/compute/vm-sample.js ? Can you please test it and make sure that it works. Lot of customers refer to this sample. I don't want the sample to be broken. Hence forth, whenever there is an update to the arm compute package, please make sure that the sample is up to date as well. It would be nice if you can add more samples.

huangpf:
Thanks @amarzavery. I just finished running it, and actually fixed an issue. Please include it for check-in/daily test runs, as it's super useful.

PR #1643, opened 2016-03-02
Make parameters optional

begoldsm:
And update the high level entry clients to have the right naming.


amarzavery:
@begoldsm - Please update from upstream autorest. I merged this PR https://github.com/Azure/azure-sdk-for-node/pull/1644 with updates to gulpfile for new code generator

PR #1646, opened 2016-03-03
Graph client update using latest swagger specs

shuagarw:


amarzavery:
@shuagarw - Please pull from upstream autorest.

PR #1647, opened 2016-03-03
regenerating datalake and improved test timing in playback mode

amarzavery:


amarzavery:
@begoldsm - Generated datalake with the new version of code generator.
Please add your mappings to gulpfile.js and stop using the .cmd files for generation.
With gulpfile you should be able to generate code crossplatform if you have mono installed on your linux box.

amarzavery:
@begoldsm - The new version has support for readonly properties. Properties that are marked as readonly will not be sent on the wire.

I have not updated the version in package.json and leave it to you to update and then use the CI to publish to npm.

PR #1668, opened 2016-03-17
Add exportTemplate api and debugSetting to deployments

vivsriaus:


amarzavery:
@vivsriaus The changes are good.
One more thing. Many customers use tests as a sample to achieve what they want to do. Do you think adding/modifying tests for the new functionality would be a good idea?

vivsriaus:
@amarzavery I thought about adding a test, but it looks like the CLI tests that I will write for the command will overlap with the sdk tests. I also didn't see a good coverage of our sdk in the existing test list. IMO, we should just keep the core positive test cases (like resource group CRUD, resource CRUD, deployment CRUD, etc) and reserve the interesting scenarios with different parameter options for each api in the CLI.

PR #1675, opened 2016-03-23
[Service Fabric] Update readme

msjeffreychen:


lib/services/serviceFabric/README.md
```diff
@@ -9,7 +9,7 @@ This project provides a Node.js package for accessing the Azure Service Fabric.
 ## How to Install
 
 ```bash
-npm install azure-servicefabric
+npm install azure-arm-servicefabric
```
amarzavery: why? you are not an ARM service and you do not use Azure.NodeJS version of the codegen. So the package name azure-servicefabric is just perfect.

msjeffreychen: We are consider ARM service, according to our PM.

amarzavery:
@msjeffreychen - Why was this done? https://www.npmjs.com/search?q=azure-servicefabric
How are those 2 packages azure-servicefabric, azure-arm-servicefabric different?
You have no reason to use the azure-arm prefix as servicefabric is not an ARM service and does not use Azure.NodeJS version of the codegen.

msjeffreychen:
The cli only supports 2 modes, arm and asm. In this case, which mode should I select?
I noticed packages are following the convention of naming arm, asm. That's why I renamed it. I will need to delete the old name.

msjeffreychen:
I just confirmed again with our PM. We are considered ARM provider, registered under the name "Microsoft.ServiceFabric". It is currently internal until we go GA. Thanks.

amarzavery:
Ok. If that is the case then I will delete the package with the name azure-servicefabric.

Please pull from Azure/master and regenerate the code. I have updated the gulpfile with today's version of the code generator. There were some bug fixes which will be helpful to you.

devigned:
@msjeffreychen the arm moniker on this spec and package should not be there. ARM is to indicate the services uses RP requests against the Azure Resource Manager API. Your service may be an ARM service, but this package does not target the ARM API (unless through some sort of a local proxy, and if that's the case, then I still think it should not be 'arm_').

PR #1676, opened 2016-03-23
Add export template resource group operation

vivsriaus:


amarzavery:
@vivsriaus  -Was the PR in rest-api-specs repo merged ?

lib/services/resourceManagement/lib/resource.js
```diff
@@ -15,7 +15,7 @@
 //
 'use strict';
 
-exports.LockManagementClient = require('./lock/lockManagementClient');
+exports.LockManagementClient = require('./lock/managementLockClient');
```
amarzavery: should this be exports.ManagementLockClient as well ?

PR #1677, opened 2016-03-24
regenerated arm network with today's version of Autorest and latest changes made to 2016-03-30 api version of the spec

amarzavery:


amarzavery:
@deepakswifty , @MikhailTryakhov, @romangromov - please review this PR for latest changes in  azure-arm-network

lib/services/networkManagement2/lib/models/expressRouteCircuitRoutesTableSummary.js
```diff
@@ -0,0 +1,86 @@
...
+ * @member {string} [neighbor] Neighbor.
+ * 
+ * @member {number} [v] V.
+ * 
+ * @member {number} [as] AS.
+ * 
+ * @member {string} [upDown] upDown.
+ * 
+ * @member {string} [statePfxRcd] statePfxRcd .
+ *
```
amarzavery: @deepakswifty - I think these properties should be better described in the swagger spec. As of now the end user cannot make any sense of this.

lib/services/networkManagement2/lib/models/expressRouteCircuitsStatsListResult.js
```diff
@@ -25,8 +27,6 @@ var util = require('util');
 function ExpressRouteCircuitsStatsListResult() {
 }
 
-util.inherits(ExpressRouteCircuitsStatsListResult, Array);
-
```
amarzavery: @deepakswifty - This change means that "x-ms-pageable-operation" extension was removed from the operation that provided expressRouteCircuitsStatsListResult as the response.

Even if the operation is not pageable, it is good to have the extension.
You can explicilty specify the nextLink to null in the extension. That will not generate the nextLink but till still provide a user friendly clientside API. 

In node.js the user will directly get the array referenced by value property a.k.a value is flattened and in C# the user will get an IEnumerable<ExpressRouteCircuitStats>.

Storage does the same thing for listing storage accounts

amarzavery: https://github.com/Azure/azure-rest-api-specs/blob/master/arm-storage/2015-06-15/swagger/storage.json#L299

DeepakRajendranMsft:
@dihan0604  please take a look at Amar's comments

PR #1681, opened 2016-03-30
FW: [SDK for node] Replace toIsoString() calls with toISOString() cal…

gucalder:
…ls and make sure moment@2.8.0 or more recent is being referenced


amarzavery:
Was this change made manually or it was autogenerated?

gucalder:
The change was done manually, and I am testing the commands with the changed file to be sure.

PR #1702, opened 2016-04-12
storage encryption & cool

emgerner-msft:


amarzavery:
@emgerner-msft - Everything else looks good. Please update the major version in package.json. I think the autorest generated storage management library has been pretty stable and we can have 1.0.0 version.

emgerner-msft:
As it stands I've changed the storage version from 0.12.2 to 0.13 -- is there any other package I have to change? Given autorest was still in preview my impression (and I believe the same for CRP and NRP which are also in preview) was that we should continue to have our RPs in preview. Are we moving autorest out of preview then?

amarzavery:
Generic ARM has moved to 1.x as you can see here. https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/resourceManagement/package.json#L9. AutoRest is also fairly stable. Whether to move autorest to 1.0 is a different conversation. 
Considering the fact that hydra generated storage management library has been there since a good time and the AutoRest generated storage management library has had few versions on top of it and we haven't had any issues in node sdk or in the CLI, I think this is stable and should be moved to 1.0.0. It is a suggestion from my side. If you think keeping this to 0.13.0 is accurate then that is fine as well.

emgerner-msft:
Ah I see. How about lets leave it for 0.13 for just this release and then start a broader discussion with CRP/NRP. SRP being the only one to move up to 1.0 isn't really helpful within itself anyways. We can at the same time discuss the GA plan for AutoRest and iron these details out. Then if CRP and NRP agree, we could actually just do a group 1.0 release with the core stuff with any last minute bug fixes. If you want to start an email thread or just add it to the discussion at one of our next internal CLI meetings (@devigned) we can close out pretty quickly. Does that sound good?

amarzavery:
sounds good to me. Merging the PR.

PR #1793, opened 2016-06-30
fixed bug where null options object prevents defaults from being set

andyroyle:
When `options` is null, then the setting for `DEFAULT_CLIENT_REQUEST_TIMEOUT` is not correctly applied (so requests will _never_ timeout).

If `options` is null, set it to an empty object.


azurecla:
Hi __@andyroyle__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #1856, opened 2016-08-23
update to latest job swagger for ADLA

begoldsm:
This is required to ensure that we properly regenerate the Node code for ADL


amarzavery:
@begoldsm - That is a known failure for 0.12.x. We shall be removing support for 0.12.x in the upcoming sprint.

Issue #1893, opened 2016-09-21
[ServiceBus] Create topic does not support all options


princjef:
I'm am using the Service Bus portion of the SDK to try to create a partitioned topic, among other things. However, [`createTopicIfNotExists()`](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/serviceBus/lib/servicebusservice.js#L754) does not allow either the SupportOrdering or EnablePartitioning options to be set.

From the [REST API docs](https://msdn.microsoft.com/en-us/library/hh780749.aspx), it looks like SupportOrdering should be available, and [this example using the C# SDK](https://azure.microsoft.com/en-us/documentation/articles/service-bus-partitioning/#_enable-partitioning) specifies the EnablePartitioning property, so it seems like that should be available too.

[This section](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/serviceBus/lib/models/topicresult.js#L24) appears to be the culprit, since it filters out any properties that aren't explicitly listed.

eduardkoller:
@sazeesha - Can you take a look at this issue? /c: @djrosanova

sazeesha:
Please refer to the Swagger [here](https://github.com/Azure/azure-rest-api-specs/blob/master/arm-servicebus/2015-08-01/swagger/servicebus.json) and use the SDK from [here](https://github.com/Azure/azure-sdk-for-net/tree/AutoRest/src/ResourceManagement/ServiceBus). This has support for SupportOrdering and EnablePartitioning.

PR #1894, opened 2016-09-21
Service Bus - Add extra options for create topic

princjef:
Addresses #1893 by adding support for SupportOrdering and EnablePartitioning to `createTopicIfNotExists()`.

It also bumps the version of the API to 2016-07, since EnablePartitioning is not supported by API version 2013-07.


azurecla:
Hi __@princjef__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla.azure.com.
    </span>

TTYL, AZPRBOT;

princjef:
I have also verified that using API version 2013-10 will work for both added options, but decided to bump to the latest because it didn't seem to break anything. I don't know what backwards compatibility between API versions looks like, but I can move the version back to 2013-10 if necessary.

PR #1912, opened 2016-09-28
update node sdk for batch management to match swagger spec

xingwu1:


package.json
```diff
@@ -72,7 +72,7 @@
     "azure-arm-website": "0.11.2",
     "azure-arm-datalake-analytics": "0.4.3",
     "azure-arm-datalake-store": "0.4.2",
-    "azure-arm-batch": "0.2.1",
+    "azure-arm-batch": "0.3.0",
```
amarzavery: instead of updating the rollup package.json please update the package.json  of the batch management library. The rollup package will be updated when the rollup version is ready for publish.

PR #1913, opened 2016-09-28
update node sdk for batch management to match swagger spec

xingwu1:


xingwu1:
@amarzavery Accidentally delete my branch, so resend a new PR, and change the version as your suggestion.

PR #1924, opened 2016-10-11
Adding provisioning state

rkmanda:
Adding provisioning state


rkmanda:
will resubmit considering the latest swagger changes

PR #1928, opened 2016-10-12
Node sdk for Analysis services 

athipp:
This is the pr for the swagger spec.
https://github.com/Azure/azure-rest-api-specs/pull/583

PR #1948, opened 2016-10-27
node SDK for azure analysis services

athipp:


amarzavery:
@athipp - You forgot to update the version in `lib/services/analysisServices/package.json`

athipp:
@amarzavery In my previous I was told to remove the version change in package.json and you had mentioned sdk team would do that - https://github.com/Azure/azure-sdk-for-node/pull/1928
Is that not the case?

PR #1981, opened 2016-11-14
Update ADL readme files, reflecting latest swagger

begoldsm:
Updating both ADL readme's prior to publish.


amarzavery:
looks good.

PR #1983, opened 2016-11-17
Adding innererror and updating api-version to 2016-10-01

pomortaz:
- Adding innererror
- Updating api-version to 2016-10-01
- Rerecording the tests and regenerating the sdk core


pomortaz:
@jeffsimmer please review.

PR #1984, opened 2016-11-17
Update batch SDK by using latest autorest and latest swagger spec.

xingwu1:


xingwu1:
@amarzavery Ping...

amarzavery:
@xingwu1 - There are 198 files here. Have you updated the package.json inside the lib/services/batch and lib/services/batchManagement folders? Does this change require, any update to the examples in the README.md files in the above two folders? Please let me know that the above things are covered and I shall merge the PR after that.

xingwu1:
@amarzavery 
1. This change only affect batch data plane, so I only updated package.json in lib/services/batch. 
2. There is no API change at all, so readme and example change.
3. I just re-ran the autorest to fix codegen bug related to nextlink. also many changes are due to use the latest swagger spec which has a lot comments change.

amarzavery:
cool sounds good. Thanks for confirming. The browser becomes super slow while reviewing 198 changed files

xingwu1:
Thank you @amarzavery

PR #1991, opened 2016-11-23
Update ARM node sdk from latest spec

vivsriaus:
Update ARM node sdk from latest spec


amarzavery:
@vivsriaus - The PR has been merged.

PR #2023, opened 2017-01-11
resourceFiles  attribute in task - add missing description

shwetams:
When we add a task and provide resource Files, the documentation mentions that it is an array of files. There is no description on the array structure which is a JSON array with three keys. The explanation text for resourceFiles can be:

"An array of files which will be downloaded and are needed for the tasks to run. Following is the structure:
[{'blobSource':<The SAS URI of the file>','filePath':'<file name and path where the file will be downloaded>','fileMode':<File mode in octal format, only applicable for Linux node, default value is 0770>}] "

This will be super useful, I personally had to struggle for sometime when I came across the .Net documentation that has a similar class and guessed the attribute names.


azurecla:
Hi __@shwetams__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (shg). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
        If you are a vendor, DO please sign the electronic contribution license agreement. It will take 2 minutes and there's no faxing! https://cla.azure.com.
    </p>

TTYL, AZPRBOT;

amarzavery:
Looks like your branch is messed up. You shouldnt be having so many commits.

PR #2039, opened 2017-01-26
updates to the runtime for supporting generic requests

amarzavery:


runtime/ms-rest/lib/webResource.js
```diff
@@ -293,4 +298,244 @@ WebResource.prototype.pipeInput = function(inputStream, destStream) {
...
+  }
+  this.method = options.method.toUpperCase();
+
+  //construct the url if path template is provided
+  if (options.pathTemplate) {
+    if (typeof options.pathTemplate !== 'string') {
+      throw new Error('options.pathTemplate must be of type \'string\'.');
+    }
+    if (!options.baseUrl) {
+      options.baseUrl = 'https://management.azure.com';
```
vishrutshah: just a thought. If this is ms-rest runtime why should we default to `https://management.azure.com` Should we be keeping ms-rest defaults to none / some http defaults and then overriding it in ms-rest-azure and using it for validation?

runtime/ms-rest/lib/webResource.js
```diff
@@ -293,4 +298,244 @@ WebResource.prototype.pipeInput = function(inputStream, destStream) {
...
+  //ensure accept-language is set correctly
+  if (!this.headers['accept-language']) {
+    this.headers['accept-language'] = 'en-US';
+  }
+  //ensure the request-id is set correctly
+  if (!this.headers['x-ms-client-request-id'] && !options.disableClientRequestId) {
+    this.headers['x-ms-client-request-id'] = utils.generateUuid();
+  }
+  //ensure content-type is set correctly
+  if (this.headers['Content-Type'] &&
```
vishrutshah: why are we adding `charset=utf-8` is user doesn't provided it? Shouldn't we respect users inputs?

runtime/ms-rest/lib/serviceClient.js
```diff
@@ -139,4 +140,98 @@ ServiceClient.prototype.addFilter = function (newFilter) {
...
+ * the object must have a property named value which provides the 'path-parameter-value'.
+ * Example: 
+ *    - path-parameter-value in 'object' format: { 'path-parameter-name': { value: 'path-parameter-value', skipUrlEncoding: true } }
+ *    - path-parameter-value in 'string' format: { 'path-parameter-name': 'path-parameter-value' }.
+ *
+ * @param {object} [options.headers] A dictionary of request headers that need to be applied to the request.
+ * Here the key is the 'header-name' and the value is the 'header-value'. The header-value MUST be of type string.
+ *  - ContentType must be provided with the key name as 'Content-Type'. Default value 'application/json; charset=utf-8'.
+ *  - 'Transfer-Encoding' is set to 'chunked' by default if 'options.bodyIsStream' is set to true.
+ *  - 'Content-Type' is set to 'application/octet-stream' by default if 'options.bodyIsStream' is set to true.
```
amarzavery: @vishrutshah 
no. By default it is application/json. However, 
> 'Content-Type' is set to 'application/octet-stream' by default **if 'options.bodyIsStream' is set to true.**

amarzavery: Line number 182

runtime/ms-rest/lib/webResource.js
```diff
@@ -293,4 +298,244 @@ WebResource.prototype.pipeInput = function(inputStream, destStream) {
...
+ */
+WebResource.prototype.validateRequestProperties = function validateRequestProperties() {
+  if (!this.method || !this.url || !this.headers['Content-Type'] || !this.headers['accept-language']) {
+    throw new Error('method, url, headers[\'Content-Type\'], headers[\'accept-language\'] are ' +
+      'required properties before making a request. Either provide them or use WebResource.prepare() method.')
+  }
+}
+
+/**
+ * Sends the request and returns the response.
```
amarzavery: done.

runtime/ms-rest/lib/webResource.js
amarzavery: we are doing this in the all the autorest generated sdks. An example over [here](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/storageManagement2/lib/operations/storageAccounts.js#L121). I think it is a standard practice.

runtime/ms-rest/lib/webResource.js
amarzavery: WebResource.js is available in ms-rest and not ms-rest-azure. It can be done, but it will be super complicated. To make it easy for us (azure users) to use the method, I have defaulted it to that specific value. Frankly speaking, the usage of this method will be very less outside the scope of azure.

PR #2073, opened 2017-02-22
Update Batch SDKs with latest REST API version

jasper-schneider:
Regenerate Batch SDKs, both management plane and data plane, based on the following new Swagger specs:
https://github.com/Azure/azure-rest-api-specs/blob/master/arm-batch/2017-01-01/swagger/BatchManagement.json
https://github.com/Azure/azure-rest-api-specs/blob/master/batch/2017-01-01.4.0/swagger/BatchService.json


package.json
```diff
@@ -73,8 +73,8 @@
     "azure-arm-website": "0.11.2",
     "azure-arm-datalake-analytics": "0.4.3",
     "azure-arm-datalake-store": "0.4.2",
-    "azure-arm-batch": "0.2.1",
-    "azure-batch": "0.5.1",
+    "azure-arm-batch": "0.4.0",
```
jasper-schneider: @amarzavery 
Just wanted to confirm: Should I be changing the top-level package.json now, or do you guys handle this after the packages have actually been published?

jasper-schneider: Well, looks like the CI job failed because it couldn't find the version specified here, so I'm going to revert my change to this file.

Issue #2076, opened 2017-02-23
Send user agent header with requests


balajikris:
Implement https://github.com/Azure/azure-sdk-for-ruby/issues/517 for node sdk

**Tasks:**
- [x] changes to runtime are merged 
- [x] updated runtime packages published to npm. 
- [x] changes to Autorest are merged. 
- [x] Regenerate all autorest generated packages and publish them to npm.

balajikris:
This is all checked in and released as of today.

PR #2077, opened 2017-02-23
send user agent info in request headers

balajikris:
These are the runtime changes required to send userAgent string in request headers. See issue #2076

This adds a `addUserAgentInfo` to the service client which is the base class for any concrete azure service client. So, autorest generated service clients or any 3rd party libraries can simply call this method and pass in their package name and module version in the format `module_name/package_version`.

The runtime then constructs the user agent string via an user agent filter as part of the request pipeline and sends it along.

### DIFF
--- a/runtime/ms-rest/lib/filters/userAgentFilter.js
+++ b/runtime/ms-rest/lib/filters/userAgentFilter.js
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+var os = require('os');
+var util = require('util');
+var Constants = require('../constants');
+var HeaderConstants = Constants.HeaderConstants;
+
+/**
+* Creates a filter to add the user agent header in a request.
+*
+* @param {string} userAgent The user agent string to use.
+*/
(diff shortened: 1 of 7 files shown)


runtime/ms-rest/lib/filters/userAgentFilter.js
```diff
@@ -0,0 +1,30 @@
...
+  return function handle(resource, next, callback) {
+    if (!resource.headers[HeaderConstants.USER_AGENT]) {
+      exports._tagRequest(resource, userAgentInfo);
+    }
+
+    return next(resource, callback);
+  };
+};
+
+exports._tagRequest = function (requestOptions, userAgentInfo) {
```
amarzavery: tagRequest

balajikris:
This [build failure] is due to a bug in our test infra. It couldn't handle a change to the dependency (ms-rest) while at the same time consuming the change (ms-rest-azure). 
I've filed workitem #2078 for this. Meanwhile, I'll try break up this change into 2 smaller changes. one for ms-rest and another for ms-rest-azure and try to get this in sequentially.

balajikris:
closing this in favor of #2079 and #2080

Issue #2078, opened 2017-02-24
Test infrastructure for ms-rest-azure should download appropriate version of ms-rest


balajikris:
this [build failure](https://travis-ci.org/Azure/azure-sdk-for-node/builds/204800899) points to a bug in our test infrastructure. If you're updating ms-rest and ms-rest-azure at the same time, where the changes in latter depend on the former, tests in ms-rest-azure may fail. This is because though the test folder in ms-rest-azure is supposed to reference the local (on disk) version of ms-rest, the infra is set up incorrectly such that it refers to latest version from npmjs.org, which will not have the changes your PR made to ms-rest.

In this case, i was adding a new function to ms-rest and invoking it in ms-rest-azure. The test wasn't able to find the definition of the function because it did not have the version of ms-rest with my changes.

Logs:

~~~
/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/lib/azureServiceClient.js:59
  this.addUserAgentInfo(util.format('%s/%s', moduleName, moduleVersion));
       ^
TypeError: this.addUserAgentInfo is not a function
    at new AzureServiceClient (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/lib/azureServiceClient.js:59:8)
    at Suite.<anonymous> (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/test/azureServiceClientTests.js:90:18)
    at context.describe.context.context (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/node_modules/mocha/lib/interfaces/bdd.js:47:10)
    at Suite.<anonymous> (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/test/azureServiceClientTests.js:32:3)
    at context.describe.context.context (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/node_modules/mocha/lib/interfaces/bdd.js:47:10)
    at Object.<anonymous> (/home/travis/build/Azure/azure-sdk-for-node/runtime/ms-rest-azure/test/azureServiceClientTests.js:15:1)
    at Module._compile (module.js:409:26)
~~~

PR #2080, opened 2017-02-24
msrestazure - send user agent info in request..

balajikris:
.. headers. Part of work for issue #2076

This is part 2 of the runtime changes. See #2078 for why runtime changes need to be made independently.


balajikris:
yep. I have to publish a new version of ms-rest and then push this change, to get CI passing.

balajikris:
Re-run tests after merging and publishing #2081

PR #2084, opened 2017-03-03
get info from package json

balajikris:
This is part of work for https://github.com/Azure/azure-sdk-for-node/issues/2076

Motivation:
To send user agent information from azure packages, we have to read its package.json. It turns out that package.json isn't always in the same relative path from the service client itself. So, this helper on the base service client helps to locate package.json and send back the required info. This will be consumed by azure clients generated from autorest.

### DIFF
--- a/runtime/ms-rest/lib/serviceClient.js
+++ b/runtime/ms-rest/lib/serviceClient.js
@@ -265,7 +265,8 @@ ServiceClient.prototype.addUserAgentInfo = function addUserAgentInfo(additionalU
  * Attempts to find package.json for the given azure nodejs package.
  * If found, returns the name and version of the package by reading the package.json
  * If package.json is not found, returns a default value.
- * @param {any} managementClientDir - pass the directory of the specific azure management client.
+ * @param {string} managementClientDir - pass the directory of the specific azure management client.
+ * @returns {object} packageJsonInfo - "name" and "version" of the desired package.
  */
 ServiceClient.prototype.getPackageJsonInfo = function getPackageJsonInfo(managementClientDir) {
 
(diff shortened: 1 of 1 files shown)


runtime/ms-rest/lib/serviceClient.js
```diff
@@ -259,4 +259,53 @@ ServiceClient.prototype.addUserAgentInfo = function addUserAgentInfo(additionalU
   this.userAgentInfo.value.push(additionalUserAgentInfo);
 };
 
+/**
+ * Attempts to find package.json for the given azure nodejs package.
+ * If found, returns the name and version of the package by reading the package.json
+ * If package.json is not found, returns a default value.
+ * @param {any} managementClientDir - pass the directory of the specific azure management client.
```
amarzavery: {string}

runtime/ms-rest/lib/serviceClient.js
```diff
@@ -259,4 +259,53 @@ ServiceClient.prototype.addUserAgentInfo = function addUserAgentInfo(additionalU
...
+  // so, package.json could be found just one level above where management client lives.
+  // In some packages (azure-arm-resource), management client lives at one level deeper in the lib directory
+  // so, we have to traverse at least two levels higher to locate package.json.
+  // The algorithm for locating package.json would then be, start at the current directory where management client lives
+  // and keep searching up until the file is located. We also limit the search depth to 2, since we know the structure of 
+  // the clients we generate.
+  //
+
+  var packageJsonInfo = {
+    name: 'default',
```
amarzavery: NO_NAME

runtime/ms-rest/lib/serviceClient.js
amarzavery: @returns {object} packageJsonInfo - "name" and "version" of the desired package

Issue #2033, opened 2017-01-22
Use promises instead of callbacks


atrauzzi:
In 2015 this ticket happened: https://github.com/Azure/azure-sdk-for-node/issues/1450

The answer was a fairly unsubstantiated _callbacks are more common_.  Even for 2015, that was just not in any way accurate.  This continues to be the case today where the `Promise` API is making its way throughout the ecosystem.  Moreover, consistency is almost universally the worst argument to uphold substandard practices.

Given that typescript has implemented `await`, it would be nice to be able to leverage that functionality instead of having to descend into [callback hell](http://callbackhell.com/).

See also:
 - https://github.com/Azure/azure-storage-node/issues/110
 - https://github.com/Azure/azure-documentdb-node/issues/154

amarzavery:
Thanks for the feedback. We are working on it. Will keep you updated.
/cc @balajikris

balajikris:
Hey @atrauzzi This is now fixed with today's update to all our SDKs in this repo. Promises are in, type definitions are refreshed and should work with the latest version of tsc.exe and much more goodness. Check our [latest release here](https://github.com/Azure/azure-sdk-for-node/releases/tag/2.0.0-preview-April2017) and feel free to let us know what you think.

/cc: @amarzavery @fiveisprime

PR #2085, opened 2017-03-04
prototype for callbackOrPromise

amarzavery:
feedback is welcome

### DIFF
--- a/lib/services/storageManagement2/lib/operations/storageAccounts.js
+++ b/lib/services/storageManagement2/lib/operations/storageAccounts.js
@@ -14,6 +14,7 @@ var util = require('util');
 var msRest = require('ms-rest');
 var msRestAzure = require('ms-rest-azure');
 var WebResource = msRest.WebResource;
+var HttpOperationResponse = msRest.HttpOperationResponse;
 
 /**
  * @class
@@ -27,6 +28,7 @@ var WebResource = msRest.WebResource;
  */
 function StorageAccounts(client) {
   this.client = client;
+  this._checkNameAvailability = _checkNameAvailability;
 }
 
 /**
@@ -39,10 +41,42 @@ function StorageAccounts(client) {
  * @param {object} [options.customHeaders] Headers that will be added to the
  * request
  *
- * @param {function} callback
+ * @returns {Promise} A promise is returned
+ * @resolve {HttpOperationResponse<CheckNameAvailabilityResult>} - The deserialized result object.
+ * @reject {Error} - The error object.
+ */
+StorageAccounts.prototype.checkNameAvailabilityWithHttpOperationResponse = function (name, options) {
+  let self = this;
+  let client = self.client;
+  return new Promise((resolve, reject) => {
+    self._checkNameAvailability(name, options, (err, result, request, response) => {
+      if (err) {
+        return reject(err);
+      }
+      let httpOperationResponse = new HttpOperationResponse(request, response);
+      httpOperationResponse.body = result;
+      return resolve(httpOperationResponse);
+    });
+  });
+};
+
+/**
+ * Checks that the storage account name is valid and is not already in use.
  *
- * @returns {function} callback(err, result, request, response)
+ * @param {string} name
  *
+ * @param {object} [options] Optional Parameters.
+ *
+ * @param {object} [options.customHeaders] Headers that will be added to the
+ * request
+ *
+ * @param {function} [optionalCallback] - 
+ * 
+ * @returns {function|Promise} If a callback was passed as the last parameter 
+ * then it returns the callback else returns a Promise.
+ * 
+ *          {function}  optionalCallback(err, result, request, response)
+ *                    
  *                      {Error}  err        - The Error object if an error occurred, null otherwise.
  *
  *                      {object} [result]   - The deserialized result object.
@@ -52,10 +86,34 @@ function StorageAccounts(client) {
  *                      {object} [request]  - The HTTP Request object if an error did not occur.
  *
  *                      {stream} [response] - The HTTP Response stream if an error did not occur.
+ * 
+ *          {Promise}   A promise is returned.
+ *             @resolve {CheckNameAvailabilityResult} - The deserialized result object.
+ *             @reject {Error} - The error object.
  */
-StorageAccounts.prototype.checkNameAvailability = function (name, options, callback) {
+StorageAccounts.prototype.checkNameAvailability = function (name, options, optionalCallback) {
   var client = this.client;
-  if(!callback && typeof options === 'function') {
+  var self = this;
+  if (!optionalCallback && typeof options === 'function') {
+    optionalCallback = options;
+    options = null;
+  }
+  if (!optionalCallback) {
+    return new Promise((resolve, reject) => {
+      self._checkNameAvailability(name, options, (err, result, request, response) => {
+        if (err) return reject(err);
+        
+        return resolve(result);
+      });
+    });
+  } else {
+    return self._checkNameAvailability(name, options, optionalCallback);
+  }
+};
+
+function _checkNameAvailability(name, options, callback) {
+  var client = this.client;
+  if (!callback && typeof options === 'function') {
     callback = options;
     options = null;
   }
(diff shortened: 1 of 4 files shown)


amarzavery:
@balajikris @fiveisprime  - Please review.

lib/services/storageManagement2/lib/operations/storageAccounts.js
```diff
@@ -52,10 +86,34 @@ function StorageAccounts(client) {
...
+        return resolve(result);
+      });
+    });
+  } else {
+    return self._checkNameAvailability(name, options, optionalCallback);
+  }
+};
+
+function _checkNameAvailability(name, options, callback) {
   var client = this.client;
```
amarzavery: I have just prorotyped for checkNameAvailability() method. All the changes from here on in this file are a result of code formatting in vscode. Please ignore them.

Do take a look at the test.js file below.

lib/services/storageManagement2/lib/operations/storageAccounts.js
```diff
@@ -52,10 +86,34 @@ function StorageAccounts(client) {
...
+        
+        return resolve(result);
+      });
+    });
+  } else {
+    return self._checkNameAvailability(name, options, optionalCallback);
+  }
+};
+
+function _checkNameAvailability(name, options, callback) {
```
amarzavery: This is the currently autogenerated method. This will become private.

lib/services/storageManagement2/lib/operations/storageAccounts.js
```diff
@@ -52,10 +86,34 @@ function StorageAccounts(client) {
  *                      {object} [request]  - The HTTP Request object if an error did not occur.
  *
  *                      {stream} [response] - The HTTP Response stream if an error did not occur.
+ * 
+ *          {Promise}   A promise is returned.
+ *             @resolve {CheckNameAvailabilityResult} - The deserialized result object.
+ *             @reject {Error} - The error object.
  */
-StorageAccounts.prototype.checkNameAvailability = function (name, options, callback) {
+StorageAccounts.prototype.checkNameAvailability = function (name, options, optionalCallback) {
```
amarzavery: The method signature remains the same. callback is now optional. If the customer does not provide a callback then we return a Promise.

lib/services/storageManagement2/lib/operations/storageAccounts.js
```diff
@@ -39,10 +41,42 @@ function StorageAccounts(client) {
  * @param {object} [options.customHeaders] Headers that will be added to the
  * request
  *
- * @param {function} callback
+ * @returns {Promise} A promise is returned
+ * @resolve {HttpOperationResponse<CheckNameAvailabilityResult>} - The deserialized result object.
+ * @reject {Error} - The error object.
+ */
+StorageAccounts.prototype.checkNameAvailabilityWithHttpOperationResponse = function (name, options) {
```
amarzavery: The callback used to provide the (err, result, request, response). If someone is interested in it then they can access them.
In Promise land, we can only return one object when resolving the Promise. Hence added this method to return a wrapper object that provides everything.

The standard method below will return a Promise that will resolve with the result (deserialized object).

amarzavery:
@atrauzzi @stevenfollis - Please review. https://github.com/Azure/azure-sdk-for-node/issues/2033

lib/services/storageManagement2/lib/operations/storageAccounts.js
```diff
@@ -39,10 +41,42 @@ function StorageAccounts(client) {
...
+  let self = this;
+  let client = self.client;
+  return new Promise((resolve, reject) => {
+    self._checkNameAvailability(name, options, (err, result, request, response) => {
+      if (err) {
+        return reject(err);
+      }
+      let httpOperationResponse = new HttpOperationResponse(request, response);
+      httpOperationResponse.body = result;
+      return resolve(httpOperationResponse);
```
atrauzzi: May want to confirm this, but I don't think `resolve` and `reject` are required to return anything and are "pure".

amarzavery: That is true.  Have the habit of returning every time from a function while coding. Hence added it. I shall remove it.

atrauzzi: I saw in another project that `return` was being used as a defensive technique inside of promises.  Not sure if that changes things here or not.

amarzavery: Well I tested the code by removing the return keyword and added a `console.log()`. The result was amazing: `console.log()` got executed.

### Modified method by removing the "return" keyword 
```javascript
StorageAccounts.prototype.checkNameAvailability = function (name, options, optionalCallback) {
  var client = this.client;
  var self = this;
  if (!optionalCallback && typeof options === 'function') {
    optionalCallback = options;
    options = null;
  }
  if (!optionalCallback) {
    return new Promise((resolve, reject) => {
      self._checkNameAvailability(name, options, (err, result, request, response) => {
        if (err) reject(err);
        console.log('Before resolving...');
        console.log(result);
        resolve(result);  //<<<<<<
        console.log('After resolving...');
      });
    });
  } else {
    return self._checkNameAvailability(name, options, optionalCallback);
  }
};
```

### The test code that called the method
```javascript
msRestAzure.loginWithServicePrincipalSecret(spn, secret, domain, function (err, credentials) {
  var client = new storageManagementClient(credentials, subscriptionId);
  let name = 'foo'; 
  //calling the method in promise style
  client.storageAccounts.checkNameAvailability(name).then(function (result) {
    console.log('In the thenable waiting for the result of the resolved promise..');
    console.log(result);
  }).catch(function (err) {
    console.log(err);
  });
});
```

##### Output:
```bash
bash-3.2$ node test.js
Before resolving...
{ nameAvailable: false,
  reason: 'AlreadyExists',
  message: 'The storage account named foo is already taken.' }
After resolving... //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
In the thenable waiting for the result of the resolved promise..
{ nameAvailable: false,
  reason: 'AlreadyExists',
  message: 'The storage account named foo is already taken.' }
```

##### Output after adding the "return" keyword back to the resolve statement:

```bash
bash-3.2$ node test.js
Before resolving...
{ nameAvailable: false,
  reason: 'AlreadyExists',
  message: 'The storage account named foo is already taken.' }
In the thenable waiting for the result of the resolved promise..
{ nameAvailable: false,
  reason: 'AlreadyExists',
  message: 'The storage account named foo is already taken.' }
bash-3.2$
```

### TLDR; return is required based on the above test.

What do you think?

amarzavery: Node version on the system is: 6.9.5

atrauzzi: That makes sense because the callback doesn't end the current function.  So yeah, it might be good evening just based on the semantics of your expectation for execution to end to keep the return statements.

stepro: The right semantics here is really to call reject or resolve as one statement and then ensure the function ends after that (whether through an explicit return or it just happens to leave the function).

```
if (err) {
    reject(err);
} else {
    resolve(result);
}
```

Or even simpler:

```
err ? reject(err) : resolve(result)
```

stepro:
Changes look good to me.

lib/services/storageManagement2/lib/operations/storageAccounts.js
atrauzzi: Agreed!

amarzavery: I see. Shall follow that.

amarzavery:
@mattchr @xingwu1 - Please review this prototype to support callbacks and promises. Please include Pablo and other team mates with JS expertise on this review. We are updating the generator to make these changes and plan to roll out these changes soon.

amarzavery:
@atrauzzi  A new version of the SDK has been published. Please check the [release page](https://github.com/Azure/azure-sdk-for-node/releases/tag/2.0.0-preview-April2017).

Issue #2091, opened 2017-03-07
User agent added recursively causing Request Too Long error!


DJBnjack:
Not quite sure why, but the new update (??) is causing the agent-header in the ms-rest module go overboard:

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Bad Request</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
<BODY><h2>Bad Request - Request Too Long</h2>
<hr><p>HTTP Error 400. The size of the request headers is too long.</p>
</BODY></HTML>
" for the default response.
    at /usr/src/app/node_modules/azure-arm-compute/lib/operations/virtualMachines.js:1166:19
    at retryCallback (/usr/src/app/node_modules/ms-rest/lib/filters/systemErrorRetryPolicyFilter.js:85:9)
    at retryCallback (/usr/src/app/node_modules/ms-rest/lib/filters/exponentialRetryPolicyFilter.js:135:9)
    at handleRedirect (/usr/src/app/node_modules/ms-rest/lib/filters/redirectFilter.js:36:9)
    at Request._callback (/usr/src/app/node_modules/ms-rest/lib/requestPipeline.js:126:16)
    at Request.self.callback (/usr/src/app/node_modules/request/request.js:187:22)
    at emitTwo (events.js:106:13)
    at Request.emit (events.js:192:7)
    at Request.<anonymous> (/usr/src/app/node_modules/request/request.js:1044:10)
    at emitOne (events.js:101:20)
  statusCode: 400,
  request:
   { rawResponse: false,
     queryString: {},
     url: 'https://management.azure.com/subscriptions/XXX/providers/Microsoft.Compute/virtualMachines?api-version=2016-04-30-preview',
     method: 'GET',
     headers:
      { 'x-ms-client-request-id': 'XXX',
        'accept-language': 'en-US',
        'Content-Type': 'application/json; charset=utf-8',
[...]

PR #2092, opened 2017-03-08
Fixes for user agent header

balajikris:
1. ensure that this filter is executed last and other user-agent
filters, if any (xplat-cli has its own) are executed first. That way,
the right user agent header is set. So we call the other filters in the
chain first and apply our function when the callstack unwinds. By now,
if another filter had set the user agent header, we wont bother setting
it.

2. check if a particular bit of useragent info is already present before
inserting it. safe fix for #2091

### DIFF
--- a/runtime/ms-rest/lib/filters/userAgentFilter.js
+++ b/runtime/ms-rest/lib/filters/userAgentFilter.js
@@ -13,17 +13,29 @@ var HeaderConstants = Constants.HeaderConstants;
 */
 exports.create = function (userAgentInfo) {
   return function handle(resource, next, callback) {
+    // This filter must be excuted last, before the request is sent over the wire.
+    // So, we call the next filter in the chain and finally when the call stack unwinds
+    // we tag the request with our payload.
+    var rest = next(resource, callback);
+    
     if (!resource.headers[HeaderConstants.USER_AGENT]) {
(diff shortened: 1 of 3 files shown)


balajikris:
d'oh, i used Array.prototype.includes() which is not supported until `6.10.0` of nodejs and has no support in 4.x LTS which I think is what we're running on.

runtime/ms-rest/lib/filters/userAgentFilter.js
```diff
@@ -13,17 +13,29 @@ var HeaderConstants = Constants.HeaderConstants;
...
 exports.tagRequest = function (requestOptions, userAgentInfo) {
-  var runtimeInfo = util.format('Node/%s', process.version);
   var osInfo = util.format('(%s-%s-%s)', os.arch(), os.type(), os.release());
-  userAgentInfo.unshift(runtimeInfo, osInfo);
+  if(userAgentInfo.indexOf(osInfo) === -1){
+    userAgentInfo.unshift(osInfo);
+  }
+
+  var runtimeInfo = util.format('Node/%s', process.version);
+  if(userAgentInfo.includes(runtimeInfo) === -1){
```
amarzavery: .indexOf

balajikris: 🤐

Issue #1935, opened 2016-10-15
Can't install azure package with yarn due to intune package.json engines value


jtbennett:
I can successfully install the azure package using npm, but when I try to install with [yarn](https://github.com/yarnpkg/yarn), I get the following error:

```
error azure-arm-intune@0.2.1: The engine "node" is incompatible with this module. Expected version "^5.0.0".
```

From [searching this repo](https://github.com/Azure/azure-sdk-for-node/search?utf8=%E2%9C%93&q=engines), it appears that the [intune package.json](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/intune/package.json) is the only one with an `engines` value specified.  Is intune really compatible with only node 5.x and won't run on any other version?  If not, please remove the engine value from package.json.

(There is a separate bug in yarn 0.15.1.  The error above occurs even if the --ignore-engines flag is used.  I'll file that issue in their repo.)

I am running:
- OSX El Capitan
- node 6.8.0
- npm 3.10.8
- yarn 0.15.1 (installed with brew)

brianfoody:
Experiencing same issue with;

- OSX El Capitan
- node 6.3.0
- npm 3.10.3
- yarn 0.16.1 (installed with brew)

thomaswmanion:
FYI, you can use `yarn --ignore-engines` to get around this problem for now.

vmakhaev:
This issue is more then 5 months old is blocking installation of Azure SDK on last 2 major NodeJS versions (6 and 7). Can we have any update on this? Are there any plans to fix it? ETA?

amarzavery:
we are gearing for a new release with new ES6 features and updates type definitions per latest version of typescript. We will be fixing this bug as well. the timeline for new release is 03/28.

pinkstrings:
Got my fingers crossed for this today!

amarzavery:
The new release is out https://github.com/Azure/azure-sdk-for-node/releases/tag/2.0.0-preview-April2017. Please feel free to reopen  the issue if the problem persists.

PR #2096, opened 2017-03-10
Update engine restriction in azure-arm-intune

pinkstrings:
Updated the engine restriction in azure-arm-intune to allow for modern versions of node (v5+) instead of only v5.

Issue: https://github.com/Azure/azure-sdk-for-node/issues/1935


amarzavery:
A new version of the SDK has been published. Please check the [release page](https://github.com/Azure/azure-sdk-for-node/releases/tag/2.0.0-preview-April2017).

PR #2107, opened 2017-03-20
Dev/xiowei/perf maintenance

vanbasten2323:
Added PerformMaintenance API. The RDFE API is called PerformMaintenance while the PS/CLI commands are called InitiateMaintenance. Also added MaintenanceStatus field in the GetDeployment response.


huangpf:
@vanbasten2323 I think you copy other teams' code into your PR as well. Please remove them and only keep the generated 'computeManagement' code.

PR #2109, opened 2017-03-21
updates to the runtime

amarzavery:
- moved runtime to es6 classes
- added support for promises to the login methods
- updated the index.d.ts files in the runtime
- moved the index.d.ts files to the top level folder in both the runtimes
- upgraded jshint to 2.9.4
- upgraded .jshintrc to support es6


amarzavery:
@balajikris @fiveisprime - Please review the changes.

runtime/ms-rest-azure/index.d.ts
```diff
@@ -241,15 +241,18 @@ export class BaseResource {
  *
```
balajikris: keep `index.d.ts` right next to `index.js` (whichever is the entry point) inside `lib`. See [TS Handbook](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) for reference.

The convention is that the TS entry point can be found in the same path as the JS entry point and thats likely where `tsc.exe` looks if we didn't specify a `typings` in `package.json`. Why do the extra effort to break it 😃

runtime/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -31,320 +31,320 @@ var moduleVersion = packageJson.version;
...
-AzureServiceClient.prototype.getLongRunningOperationResult = function (resultOfInitialRequest, options, callback) {
-  var self = this;
-  
-  if (!callback && typeof options === 'function') {
-    callback = options;
-    options = null;
-  }
-  if (!callback) {
-    throw new Error('Missing callback');
+    this.addUserAgentInfo(util.format('%s/%s', moduleName, moduleVersion));
```
balajikris: Since you're using ES6 throughout, this is a good candidate too. Use `interpolated` or `template` strings and even see if we can lose the dependency on `util`

balajikris: I see more usages below. So, if we're doing this, do it for all occurrences..

runtime/ms-rest-azure/lib/azureServiceClient.js
```diff
@@ -31,320 +31,320 @@ var moduleVersion = packageJson.version;
...
+
+  _checkResponseStatusCodeFailed(initialRequest) {
+    let statusCode = initialRequest.response.statusCode;
+    let method = initialRequest.request.method;
+    if (statusCode === 200 || statusCode === 202 ||
+      (statusCode === 201 && method === 'PUT') ||
+      (statusCode === 204 && (method === 'DELETE' || method === 'POST'))) {
+      return false;
+    } else {
+      return true;
```
balajikris: this whole block can be refactored into `return ! (condition)`

runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -17,163 +15,161 @@ var util = require('util');
...
-    this.details.forEach(function (element1) {
-      if (element1) {
-        element1 = this.serialize(element1);
-      }
-      deserializedArray.push(element1);
-    });
-    payload.error.details = deserializedArray;
+    if (this.details) {
+      var deserializedArray = [];
+      this.details.forEach(function (element1) {
```
balajikris: arrow syntax ?

runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -17,163 +15,161 @@ var util = require('util');
...
-        instance.error.details.forEach(function (element1) {
-          if (element1) {
-            element1 = this.deserialize(element1);
-          }
-          deserializedArray.push(element1);
-        });
-        this.details = deserializedArray;
+        if (instance.error.details) {
+          var deserializedArray = [];
+          instance.error.details.forEach(function (element1) {
```
balajikris: arrow syntax?

runtime/ms-rest/index.d.ts
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * @property {string} userName                 User name.
+ * @property {string} password                 Password.
+ * @property {string} [authorizationScheme]    The authorization scheme. Default ('Basic')
+ */
+export interface BasicAuthenticationCredentials extends ServiceClientCredentials {
+  userName: string,
+  password: string,
+  authorizationScheme?: string
+}
```
balajikris: include this in a small typescript app and ensure `tsc.exe` doesn't report any errors

balajikris:
Took a cursory look and had some suggestions. The diffs are quite large in many files, will take a in-depth look later.

runtime/ms-rest-azure/index.d.ts
amarzavery: As per the handbook reference
Also note that if your main declaration file is named index.d.ts and **lives at the root of the package **(next to index.js) you do not need to mark the "types" property, _though it is advisable to do so_.

amarzavery: Previously index.d.ts was underneath the lib directory so based on the above statement brought it at the root and also added types to package.json (as per their advise).

runtime/ms-rest-azure/index.d.ts
```diff
@@ -285,12 +304,15 @@ export function loginWithUsernamePassword(username: string, password: string, op
...
  *                      {Error}  [err]                                - The Error object if an error occurred, null otherwise.
- *
  *                      {ApplicationTokenCredentials} [credentials]   - The ApplicationTokenCredentials object
+ *     {Promise} A promise is returned.
+ *             @resolve {ApplicationTokenCredentials} The ApplicationTokenCredentials object.
+ *             @reject {Error} - The error object.
  */
-export function loginWithServicePrincipalSecret(clientId: string, secret: string, domain: string, options: AzureTokenCredentialsOptions, callback: { (err: Error, credentials: ApplicationTokenCredentials): void }): void;
\ No newline at end of file
+export function loginWithServicePrincipalSecret(clientId: string, secret: string, domain: string, options?: AzureTokenCredentialsOptions, optionalCallback?: { (err: Error, credentials: ApplicationTokenCredentials): void }): void | Promise<ApplicationTokenCredentials>;
```
lostintangent: Yay! Promise support is much appreciated :)

lib/serviceruntime/roleenvironment.js
```diff
@@ -470,6 +459,17 @@ RoleEnvironment._initialize = function (callback) {
...
     });
   };
+  
+  var getCurrentGoalState = function (finalCallback) {
+    RoleEnvironment.runtimeClient.getCurrentGoalState(function (error, goalState) {
+      if (error) {
+        callback(error);
+      } else {
+        currentGoalState = goalState;
+        getCurrentEnvironmentData(goalState, finalCallback);
```
amarzavery: moved this function after it's definition of `getCurrentEnvironmentData` to comply with jshint

runtime/ms-rest-azure/index.d.ts
```diff
@@ -26,6 +36,9 @@ export class AzureServiceClient extends msRest.ServiceClient {
     * Long Running Operations. Default value is 30.
   */
   constructor(credentials: msRest.ServiceClientCredentials, options: AzureServiceClientOptions)
+  getLongRunningOperationResult(resultOfInitialRequest: msRest.HttpOperationResponse<T>, options?: LongRunningUrlBasedRequestPrepareOptions | LongRunningPathTemplateBasedRequestPrepareOptions, optionalCallback?: msRest.ServiceCallback<TResult>): void | Promise<TResult>;
+  sendLongRunningRequest(options: LongRunningUrlBasedRequestPrepareOptions | LongRunningPathTemplateBasedRequestPrepareOptions, optionalCallack?: msRest.ServiceCallback<TResult>): void | Promise<TResult>;
+  sendLongRunningRequestWithHttpOperationResponse(options: LongRunningUrlBasedRequestPrepareOptions | LongRunningPathTemplateBasedRequestPrepareOptions): Promise<msRest.HttpOperationResponse<T>>;
```
amarzavery: These 3 methods enables a user to manually send a request that is long running and the method will automatically poll.

PR #2136, opened 2017-04-17
KeyVaultCredentials type definition fixes: (1) constructor keyword mi…

mszlapa:
KeyVaultCredentials type definition fixes: (1) constructor keyword misspelled, (2) type parameters not matching expected runtime parameter

This is a rather critical build stopping issue for us, we are not able to 'patch' the class type definition in our code, as  class merging does not work same way as interface merging.

### DIFF
--- a/lib/services/keyVault/lib/keyvault.d.ts
+++ b/lib/services/keyVault/lib/keyvault.d.ts
@@ -42,7 +42,7 @@ export interface Authenticator {
  * @param {KeyVaultCredentials~authRequest} authenticator  A callback that receives a challenge and returns an authentication token.
  */
 export class KeyVaultCredentials implements msRest.ServiceClientCredentials {
-  cosntructor(authenticator: Authenticator);
+  constructor( authenticator:  (challenge: any, callback: any) => any );
   signRequest(webResource: msRest.WebResource, callback: { (err: Error): void }): void;
 }
 


lib/services/keyVault/lib/keyvault.d.ts
```diff
@@ -42,7 +42,7 @@ export interface Authenticator {
  * @param {KeyVaultCredentials~authRequest} authenticator  A callback that receives a challenge and returns an authentication token.
  */
 export class KeyVaultCredentials implements msRest.ServiceClientCredentials {
-  cosntructor(authenticator: Authenticator);
+  constructor( authenticator:  (challenge: any, callback: any) => any );
```
amarzavery: while I agree with the constructor typo. Why would the Authenticator type not work? `constructor(authenticator: Authenticator);` Is there a reason to make challenge and callback of type "any"?
We have exposed the interface Authenticator [here](https://github.com/Azure/azure-sdk-for-node/pull/2136/files#diff-909a9a4b181a56c07f397c89bd074be5R31).
/cc @balajikris - We need to ship a patch version (with constructor fix) irrespective of the TS samples.

balajikris:
I have local changes as well. I'll look at this PR and take it through publishing

mszlapa:
KeyVaultCredentials expects a function (not an object) as a constructor parameter:

https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/keyVault/lib/keyVaultCredentials.js#L64

balajikris:
I think @mszlapa is right about the shape of the authenticator. I see a "authenticator is not a function" error if i instantiate and pass an `Authenticator` type defined in the .d.ts.

lib/services/keyVault/lib/keyvault.d.ts
balajikris: it turns out this was about the best signature I could come up with as well. At the maximum, I was able to strongly type `challenge`, which makes it no more useful than this. Also,  this `authenticator` function needs `adal-node` library, which itself is untyped at the moment. For these reasons, I agree that `(any, any) => any` is the best we could do now. We can revisit this in the future, if needed.

balajikris:
thanks @mszlapa ! will publish an update to this library in a bit.

balajikris:
@mszlapa : published `2.0.1-preview` to npm. hopefully this unblocks you. If you find issues, please get in touch and thank you for the fix 🥇

PR #2148, opened 2017-04-27
corrected import stmt names in ts file to match dataLakeStore.js

aardvarkads:
This fixes typescript compile issue due to naming mismatch between d.ts and associated js files.

Changed the property names in the dataLakeStore.d.ts file to match the associated dataLakeStore.js file

### DIFF
--- a/lib/services/dataLake.Store/lib/datalakeStore.d.ts
+++ b/lib/services/dataLake.Store/lib/datalakeStore.d.ts
@@ -4,11 +4,11 @@
  * license information.
  */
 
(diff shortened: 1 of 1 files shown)


balajikris:
@aardvarkads - Thanks for figuring out and fixing this! Good catch!

@amarzavery - I'm trying to understand if 
1. you hand-wrote this file (in which case, I can merge this PR)
or
2. if this file is auto-generated (in which case, the fix should be in the [Templates](https://github.com/Azure/autorest/tree/master/src/generator/AutoRest.NodeJS/Templates))

amarzavery:
@balajikris - This was handwritten. Please merge and publish the package with a patch version increment.

PR #2153, opened 2017-05-03
RoutingProperties changes for NodeSDK

ssankar1984:
Swagger was recently changed to include routing https://github.com/Azure/azure-rest-api-specs/pull/1174 
The following PR incorporates those changes to node sdk

PR #2154, opened 2017-05-03
Regenerate code for iothub after spec change in

balajikris:
https://github.com/Azure/azure-rest-api-specs/pull/1174


balajikris:
Submitting on behalf of @ssankar1984 (original PR: #2153)

PR #2164, opened 2017-05-11
Giving uniformity to the documentation

JJ:
Just nitpicking as is usual with professors :-)


balajikris:
Looks better now, Thanks @JJ !

fiveisprime:
👋 @JJ 

Nice to see you here and thanks for the contribution!

JJ:
I will try not to make this the last one :-) Thanks for accepting it.

PR #2165, opened 2017-05-12
Generating new package with the api version change for graphrbac

darshanhs90:
updated the package.json to use the changes from the swagger specs PR
https://github.com/Azure/azure-rest-api-specs/pull/1103


lib/services/graphManagement/package.json
```diff
@@ -6,7 +6,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Agarwal, Shubham <shuagarw@microsoft.com>"
```
wisexp: >Agarwal, Shubham <shuagarw@microsoft.com> [](start = 5, length = 41)

do we need to update this field as well?

PR #2173, opened 2017-05-22
add rdbms preview clients

krancour:


gulpfile.js
```diff
@@ -219,6 +219,16 @@ const mappings = {
     'dir': 'powerbiembedded/lib',
     'source': 'arm-powerbiembedded/2016-01-29/swagger/powerbiembedded.json'
   },
+  'rdbms.mysql': {
+    'packageName': 'azure-arm-rdbms',
```
balajikris: >rdbms [](start = 30, length = 5)

'azure-arm-mysql'. This is the name of the package you'll be publishing to NPM.

gulpfile.js
```diff
@@ -219,6 +219,16 @@ const mappings = {
...
+    'dir': 'rdbmsManagement/lib/mysql',
+    'source': 'arm-rdbms/2017-04-30-preview/swagger/Rdms.MySQL.json'
+  },
+  'rdbms.postgresql': {
+    'packageName': 'azure-arm-rdbms',
```
balajikris: >'azure-arm-rdbms [](start = 19, length = 16)

similarly, 'azure-arm-postgres' or 'azure-arm-postgresql' whichever suits better.

gulpfile.js
```diff
@@ -219,6 +219,16 @@ const mappings = {
     'dir': 'powerbiembedded/lib',
     'source': 'arm-powerbiembedded/2016-01-29/swagger/powerbiembedded.json'
   },
+  'rdbms.mysql': {
+    'packageName': 'azure-arm-rdbms',
+    'dir': 'rdbmsManagement/lib/mysql',
+    'source': 'arm-rdbms/2017-04-30-preview/swagger/Rdms.MySQL.json'
+  },
```
balajikris: I'm not sure about this. It looks like `arm-rdbms` is checked into a private repo and is still not ready for the public repo `azure-rest-api-specs`.  I'm not sure if this repo hosts SDKs generated from private specs repository. Technically, I do not see a challenge as to why we cannot, but I'm not clear if that is the norm. @amarzavery ?

gulpfile.js
balajikris: Oh I see. you want to generate a single SDK entrypoint for both these services. In that case, this is correct.

---
In reply to: [118064872](https://github.com/Azure/azure-sdk-for-node/pull/2173#discussion_r118064872) [](ancestors = 118064872)

balajikris:
@krancour - Since the PR is a mix of generated code and handwritten code, could you please give me a list of files that were not machine generated? I'd like to pay closer attention to those to ensure both languages (typescript and javascript) work as intended.

lib/services/rdbmsManagement/lib/rdbmsManagementClient.d.ts
```diff
@@ -0,0 +1,13 @@
...
+ * license information.
+ */
+
+import MySQLManagementClient = require('./mysql/mySQLManagementClient');
+import * as MySQLModels from './mysql/models';
+
+import PostgreSQLManagementClient = require('./postgresql/postgreSQLManagementClient');
+import * as PostgreSQLModels from './postgresql/models';
+
+export { MySQLManagementClient, MySQLModels, PostgreSQLManagementClient, PostgreSQLModels };
```
balajikris: Neat ! I assume this was hand-written? 

Would you please do us a favor - open a terminal, cd to this path and run `tsc.exe` on this `.d.ts` file and ensure it compiles with 0 errors and 0 warnings. We're still working on having a good infrastructure to ensure TS support works, hence the short term requirement to at least execute `tsc.exe` by hand.

balajikris:
Since this is a brand new package, there would be no tests at all. you have to add tests in here `azure-sdk-for-node\test\services\rdbms\`. you may have to touch these files `azure-tests.js`, `testlist.txt`, `testlistarm.txt` at path `azure-sdk-for-node\test\` so that the CI picks them up automatically.

lib/services/rdbmsManagement/README.md
```diff
@@ -0,0 +1,304 @@
+# Microsoft Azure SDK for Node.js - RDBMS Management
+
+This project provides a Node.js package that makes it easy to manage Microsoft Azure Storage Resources.Right now it supports:
```
balajikris: nitpick: space after period

lib/services/rdbmsManagement/README.md
```diff
@@ -0,0 +1,304 @@
...
+
+## How to Use
+
+### PostgreSQL
+
+#### Listing PostgreSQL servers
+
+```javascript
+const msRestAzure = require('ms-rest-azure');
+const PostgreSQLManagementClient = require('azure-arm-rdbms').PostgreSQLManagementClient;
```
balajikris: >require('azure-arm-rdbms').PostgreSQLManagementClient [](start = 35, length = 53)

Note that this is a side effect of exporting both 'postgres' and 'mysql' from the same package. Javascript programmers typically prefer to do `var postgresClient = require(packagename)` . The extra `.FooBarClient` is perceived a slight discomfort in the community, for reasons I'm not fully aware of. I would imagine that there would be no intellisense off the dot to list the available clients, but I believe the reasons for this goes beyond intellisense.

That said, it is entirely up to you to decide how you'd like to ship and maintain these two service clients - as two packages or as one :)

lib/services/rdbmsManagement/README.md
```diff
@@ -0,0 +1,304 @@
...
+    }
+  });
+}).then((server) => {
+  console.log('Server:');
+  console.dir(server, {depth: null, colors: true});
+}).catch((err) => {
+  console.log('An error ocurred');
+  console.dir(err, {depth: null, colors: true});
+});
+```
```
balajikris: Great set of examples on how to use the package. Even better with the usage of ES6 syntax, esp. with promises instead of callbacks :+1: 

My suggestion is, if possible, please add some typescript examples as well. Reasons:
1. This package is shipping with .d.ts files so typescript support is expected to be fully functional
2. Import syntax are better and different in typescript but there are also many ways to import. That depends on how your package exports things. So having examples would be great for this.
3. Since we are on Node 6.x, async/await is not fully supported yet. But Typescript fully supports async/await and compiles it down to whatever version of javascript the user prefers. This would make the promises chain code look even simpler.

It may not have to be as exhaustive as you have for `javascript` but having atleast a few would help :)

gulpfile.js
balajikris: I think this file (gulp task) will execute on `azure-rest-api-specs` repo. So, if someone (like my team) wants to generate all sdks to publish a refresh (we do language updates for eg.), this script will break because this spec is not in the public repo. I think the right thing to do here is to move your spec to the public repo

---
In reply to: [118066968](https://github.com/Azure/azure-sdk-for-node/pull/2173#discussion_r118066968) [](ancestors = 118066968)

gulpfile.js
krancour: Yes. afaik, there's a single resource provider that handles both of these, so this seems the most correct approach.

gulpfile.js
krancour: Agree with you. I can see where that will complicate any effort at regenerating all. I saw the Python SDK has an autogenerated client for these new services already, generated from the private repo and though it would be ok. But again... I totally see your point. Do you know if it's actually ok for us to go about moving that swagger spec to the public repo? Or do you know who we can check with?

krancour:
> Since the PR is a mix of generated code and handwritten code, could you please give me a list of files that were not machine generated?

@balajikris That's a very reasonable request. How about I just separate this into two commits where one has the auto-generated code and the other had anything that was written by hand?

krancour:
@balajikris I'be broken this up into two modules instead of just one. I've also broken the PR into four commits- handcrafted code and autogenerated code as separate commits for each of PostgreSQL and for MySQL. I know you were pleased about the typescript support-- that was actually generated by autorest. Maybe that's a new-ish feature?

I am still working toward adding tests. Typescript examples will be harder to furnish since I'm not yet very familiar with it. I may consult a colleague on that.

Thanks again for the help and reviews.

balajikris:
Hey @krancour 

> broken this up into two modules instead of just one.

I'm interested to know what compelled you to go this route. I'm definitely not complaining, but simply trying to understand the rationale. I thought you made a good argument about how they were provided by the same service provider and so the logical grouping made sense.

> typescript support

yep, we built it into autorest, but still in some cases it requires a little bit of work from the developer, to put in the typings entry point. For e.g if the javascript entry point was hand-written, the TS entry point also needs to be handwritten. It also requires some thinking to make sure the design of exports looks alright from the consumer's perspective and most importantly of all - things actually work. I thought this PR was quite diligent about that 😉 

> I am still working toward adding tests.

Thanks! feel free to leverage the spec owner's support to see if they already have something somewhere else. 

> Typescript examples will be harder to furnish since I'm not yet very familiar with it

No problem! Typescript syntax is much like C# 's and it compiles down to a given version of javascript as specified by the user. Only you get a bit of type checking for javascript programs with it. If you need an example, feel free to checkout a sample I have [here](https://github.com/balajikris/keyvault-management-typescript/blob/master/src/index.ts).  I wouldn't block check-in on this, if you do not have time.

gulpfile.js
```diff
@@ -214,6 +214,11 @@ const mappings = {
     'modeler': 'CompositeSwagger',
     'clientName': 'OperationalInsightsManagementClient'
   },
+  'postgresql': {
+    'packageName': 'azure-arm-postgresql',
+    'dir': 'postgresqlManagement/lib',
+    'source': 'arm-rdbms/2017-04-30-preview/swagger/Rdms.PostgreSQL.json'
```
balajikris: Note: There is one other parameter that could be in play here. You'd notice in some other places in this file (e.g. line 242), there is a parameter called 'FT'. Its called flattening threshold and its passed to autorest for code gen. What it controls is the signature of the generated APIs. notice that most APIs take an `options` object which is a property bag. But if there was only one property inside of it, say for e.g: `serverName`, then the API would be much cleaner if we surfaced `serverName` to the top rather than having an `options` object with 1 property inside it. 

There is no rule of thumb that dictates what FT one should be using, its more of an art, where we look at the generated API and see if it makes sense to flatten some objects. As a consumer of this API, you may discover this need when you're programming against it.

For what its worth, I looked at the shape of the generated API and it looks good as is to me. So we don't yet have to think about flattening and can revisit it later if required.

krancour:
>> broken this up into two modules instead of just one.

> I'm interested to know what compelled you to go this route.

The more I have played with this (crafting tests, etc.) I've become convinced that even if both services are implemented by a single resource provider (which, I'm no longer actually certain of, btw) they probably belong as separate modules. Two compelling things drove me to that:

1. I'm not a super experience JS guy. I'm more of a Rubyist and Gopher. So, I'm taking your word for it that, as you indicated previously, the JS community frowns upon the following:

    ```
    require('azure-arm-rdbms').PostgreSQLManagementClient
    ```

    The most obvious way to avoid that seemed to be by splitting these two clients into two separate modules.

2. In all probability, use of the PostgreSQL client or MySQL client is probably an either-or (but not both) proposition. It doesn't make a lot of sense to have a single module provide both and thus have the one you're probably _not_ using occupying memory quite unnecessarily.

krancour:
@balajikris I have added tests. I think maybe this can be re-reviewed now. Thanks again for all your help with this.

balajikris:
@krancour - Thanks, everything looks great! Just one more question - the tests you added, are they live tests ? I think so, as i didn't see any recordings or mocks. asking just for my clarification.

balajikris:
@krancour - I've merged this PR. Big thank you for the contribution! ⭐️ 

I assume you'd like to now consume these packages from NPM ? If so, you need to publish it to NPM from our official npm account. 
Doing so is easy, just put in your package name [here ](http://azuresdkci.cloudapp.net/view/5-Signing%20and%20Publishing/job/sdk-for-node-publish/build?delay=0sec) and hit build. You can do this yourself, or I can do this if you're unable to access this page or are having difficulties with it.

krancour:
@balajikris thanks for merging. Sorry for my absence the past couple days. My 10 month old caught some stomach bug and I've had to take the last couple days off to take care of him.

I won't publish the modules until we have run this past the spec / RP owner(s).

Also, to close the loop on one more detail:

> Thanks, everything looks great! Just one more question - the tests you added, are they live tests ? I think so, as i didn't see any recordings or mocks. asking just for my clarification.

There were test recordings included with the commits containing generated code.

balajikris:
No worries @krancour . Hope things are okay with you. We'll take this from here and take it through to publishing, we'll keep you in the loop as we do it. Thanks again for the contribution 👍

PR #2175, opened 2017-05-22
[ms-rest] Using \W instead of \w for regex match to include some characters

### LINKED ISSUE #2174: [ms-rest] WebResource Preparer does not handle pathParameters when it contains -
Given options as 

```
{
  "pathTemplate": "/storage/{storage-account-name}"
  "pathParameters": {
     "storage-account-name": "abcd"
   }
}
```

does not resolve the path parameters correctly. 

Related Issue: https://github.com/Azure/oav/issues/119

**cc:**
@balajikris @amarzavery

vishrutshah:
fixes #2174

### DIFF
--- a/runtime/ms-rest/lib/webResource.js
+++ b/runtime/ms-rest/lib/webResource.js
@@ -163,7 +163,7 @@ class WebResource {
       }
       let baseUrl = options.baseUrl;
       let url = baseUrl + (baseUrl.endsWith('/') ? '' : '/') + (options.pathTemplate.startsWith('/') ? options.pathTemplate.slice(1) : options.pathTemplate);
-      let segments = url.match(/({\w*\s*\w*})/ig);
+      let segments = url.match(/({\W*\S*\W*})/ig);
       if (segments && segments.length) {
         if (options.pathParameters === null || options.pathParameters === undefined || typeof options.pathParameters !== 'object') {
           throw new Error(`pathTemplate: ${options.pathTemplate} has been provided. Hence, options.pathParameters ` +


runtime/ms-rest/lib/webResource.js
```diff
@@ -163,7 +163,7 @@ class WebResource {
       }
       let baseUrl = options.baseUrl;
       let url = baseUrl + (baseUrl.endsWith('/') ? '' : '/') + (options.pathTemplate.startsWith('/') ? options.pathTemplate.slice(1) : options.pathTemplate);
-      let segments = url.match(/({\w*\s*\w*})/ig);
+      let segments = url.match(/({\W*\S*\W*})/ig);
```
balajikris: I'm not sure I understand the fix. With this change, a text like `{foo bar}` that used to be a match for the old pattern, will no longer be a match. Is that intentional, why?

In all, this change can be summarized as so :
1. change looking for any word character to any non word character
2. change looking for a space to any non-whitespace character

I don't understand this change, feel free to drop by and talk to me 😃

veronicagg: Yeah, not sure I understand it either. Is "-" the only special char we want to allow or do we allow others? if so, should we just have the expression accept "-" between words? 
I'm not sure about the original regex either though, do we allow spaces in between words for the property inside {}?

vishrutshah: @balajikris @veronicagg Yes those are valid questions. I am also not sure why the initial regex was allowing `white space` / `tabs` between `{}`. 

Secondly, after looking around I saw that technically [swagger path templating](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#path-templating) & [rfc6570](https://tools.ietf.org/html/rfc6570#section-2.3) does not say that `-` or `space` is not allowed. Moreover, [sway](https://github.com/apigee-127/swagger-tools/issues/289) and it's [regex matcher](https://github.com/pillarjs/path-to-regexp/issues/63) intentionally decided not to support `-`. 

The question remains for us is, path [/storage/{storage-account-name}](https://github.com/Azure/azure-rest-api-specs/blob/master/keyvault/2016-10-01/swagger/keyvault.json#L2390) seems to be valid in `keyvault/2016-10-01/swagger/keyvault.json` with parameter defined as ["name": "storage-account-name"](https://github.com/Azure/azure-rest-api-specs/blob/master/keyvault/2016-10-01/swagger/keyvault.json#L2399). The swagger validation tool does not complain about semantic validation but 
1. From linter perspective, do we have any reason for this to be camelCased? For example similar to Rule [M3016](https://github.com/Azure/autorest/blob/master/docs/developer/validation-rules/camel-casing-validation-rules.md)
 2. From model validation perspective how to approach this?

Let me know your thoughts!!

PR #2177, opened 2017-05-24
Update Node SDK from latest Compute spec PR

huangpf:
https://github.com/Azure/azure-rest-api-specs/pull/1238


balajikris:
Can you give me a list of files that were hand-written, if any? Or separate the commit into two, one for hand-written files and one for generated files, so that reviewing is easier.

huangpf:
Hi @balajikris , this is Node SDK PR, so it's all generated. Thanks.

lib/services/computeManagement2/package.json
```diff
@@ -12,7 +12,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Sun, Carissa <caris@microsoft.com>"
   ],
-  "version": "1.0.0-preview",
+  "version": "2.0.0-preview",
```
balajikris: do you require actually require a major version update? If the changes are not breaking and are incremental, you could just increment the minor version. That way, people who use semantic versioning `^1.0.0` would also get your update.

huangpf: The API version is upgraded from ```preview``` to non-```preview```, and there is one property field that is deprecated. Does it qualify?

balajikris: 1. then it makes sense drop the `-preview` for the package version as well?
2. the diff certainly is a lot more than deprecating 1 field. may be the previous changes to spec weren't generated and checked in to the sdk. given that, i don't mind you sticking with `2.0.0` versioning :)

huangpf: For 2., yes there are new feature changes as well, but I don't think they are the decisive factors for major version upgrade.

For 1., what's the condition that the ```-preview``` suffix on the client version string can be removed, and what future conditions do we need to keep? If it's not clear, we tend not to remove it yet.

balajikris: NPM uses semantic versioning in the form of `major.minor.patch`. When your package is in preview, I suppose it has a major version of `0` . Once your package graduates to use a major version, it is no longer in preview. I never understood the usage of `-preview` in versioning. If you're specs out of preview, then I thought the package should reflect that. No?

huangpf: The API version mostly reflects the service side change and promise, and I'm not sure about the client side requirement of removing the ```-preview``` suffix yet. So, I keep it as-is. Is it OK?

balajikris: No problem 👍

balajikris:
@huangpf - please update the title and remove "Do not merge" whenever you think this is ready to merge. I'll take over from there. Thanks

huangpf:
Thanks @balajikris . We're pending on the Swagger spec PR (Azure/azure-rest-api-specs#1238). Just saying, but if the GitHub merge can auto-depend on another GitHub PR, that would be saving us a lot of pausing and resuming time...

Issue #2191, opened 2017-06-06
Documentation Updates


fiveisprime:
- [x] Fix the [cognitive services list](https://github.com/MicrosoftDocs/azure-docs-sdk-node/blob/master/docs-ref-autogen/azure-arm-cognitiveservices/azure-arm-cognitiveservices.The-list-of-cognitive-services-accounts-operation-response..yml)
- [x] Create service mapping file (see [python](https://github.com/Azure/azure-sdk-for-python/blob/master/package_service_mapping.json) for an example)
- [x] Remove JSDoc `@class`
- [x] Place description/summary before `@constructor`

fiveisprime:
Most of these are either handled already or being handled on the other side. I'm going to go ahead and close this.

PR #2192, opened 2017-06-07
Add package service mapping file.

fiveisprime:
Required file for documentation generation. I Actually scripted the creation of
this so please review it and let me know if anything is obviously borked.

References #2191

### DIFF
--- a/package_service_mapping.json
+++ b/package_service_mapping.json
@@ -1,0 +1,278 @@
+{
+  "azure-arm-advisor": {
+    "service_name": "advisor",
+    "category": "Management"
+  },
+  "azure-arm-analysisservices": {
+    "service_name": "analysisservices",
+    "category": "Management"
+  },
+  "azure-arm-authorization": {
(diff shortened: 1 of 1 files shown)


fiveisprime:
@bradygaster ☝️

fiveisprime:
1. I can't add a header, it's JSON so no comments allowed
2. I'm not sure where to add this, but you're right (also see 3)
3. Right repo (following the python example), I think this should be generated from a gulp task and kept up to date automatically
4. See 3 and yes, exactly. I believe this particular check-in to be a stop-gap to unblock the docs team with some technical debt to get this automated

:)

balajikris:
1. we can put the how-to in a ReadME (developer section) or a Developer ReadMe if one is appropriate.
1. cool, lets have it as is for now and look at integrating it into the gulpfile as one of the tasks. 🤖

fiveisprime:
Ping @balajikris since I don't have merge access 👌

PR #2200, opened 2017-06-12
Update service mapping.

fiveisprime:
Sorted services alphabetically, cleaned up the names, and removed the
data plane pieces.


fiveisprime:
Quick ping 🏓 on this. I don't have merge access. :)

PR #2207, opened 2017-06-20
Change VMs to Virtual Machines

bradygaster:


bradygaster:
@fiveisprime spelling out virtual machines so we get this in the UX for docs. let me know if this needs to go to the jsdoc branch instead.

balajikris:
if the docs team is using `jsdocs` branch as their source, then I think this should goto that branch as having this in master would have no impact for the docs team?

PR #2217, opened 2017-07-06
Regenerated NRP to include latest features

EvgenyAgafonchikov:


DeepakRajendranMsft:
@balajikris Could you please take a look at this PR and merge, so @EvgenyAgafonchikov can send the PR the cli asap

lib/services/networkManagement2/package.json
```diff
@@ -6,7 +6,7 @@
     "Rajendran, Deepak <derajen@microsoft.com>",
     "Choubey, Ankur <anchoube@microsoft.com>"
   ],
-  "version": "1.3.0-preview",
+  "version": "1.3.1-preview",
```
balajikris: you're changing the API version but you're only updating the patch version for the package?

balajikris:
This should be closed in favor of #2219. Reopen if thats not the case.

PR #2219, opened 2017-07-07
[Network] Regenerated Network SDK using the latest swagger

number213:
This PR also contains changes from https://github.com/Azure/azure-sdk-for-node/pull/2217


balajikris:
If you have changes that conflict with another PR, then the right way to go about it, is that you have 2 commits. Pick the first commit from the PR that has the changes you need and then make your commit on top of it. That way you can avoid merge conflicts and git will handle the rest.

number213:
@balajikris this PR doesn't really depend on another one, it's that it uses more recent version of swagger.

balajikris:
are you saying if this is merged, the other PR can be closed?

number213:
@balajikris yes

lib/services/networkManagement2/package.json
```diff
@@ -6,7 +6,7 @@
     "Rajendran, Deepak <derajen@microsoft.com>",
     "Choubey, Ankur <anchoube@microsoft.com>"
   ],
-  "version": "1.3.0-preview",
+  "version": "1.4.0-preview",
```
balajikris: this describes that the changes are "minor" and non-breaking. Is that really the case? asking because you seem to have api-version update.

balajikris: Semver's guidance is this:
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards-compatible manner, and
PATCH version when you make backwards-compatible bug fixes.

number213: I ran test with new API and everything worked except recent `app-gw ssl-policy list-available` command due to SDK method being renamed. I addressed this issue in Xplat CLI PR that's going to use this version of SDK.
Should this be considered breaking change?

balajikris: yep, if a method that was shipped was renamed, then this is definitely a breaking change. 

How? Because the consumers of that method will be broken if they upgrade to the newer package. To keep their applications from being broken, they'd like their app to stick to a compatible version of the package. If we didn't follow semver's guidelines [rule 1 shown above applies here], we'd be breaking them.

balajikris: @DeepakRajendranMsft - I think you should weigh in on what changes this API brings, rather than Anton and me guessing if this is a breaking change or not 😃

number213:
@balajikris because this PR introduces breaking changes, I updated version to 2.0.0

PR #2223, opened 2017-07-19
mediaServicesManagement: correct package.json with filename case sens…

akshatcm:
…itivity

Corrected filenames in package.json. 'mediaservicesManagementClient' is changed to 'mediaservicesManagementClient'.

### DIFF
--- a/lib/services/mediaServicesManagement/package.json
+++ b/lib/services/mediaServicesManagement/package.json
@@ -4,7 +4,7 @@
   "contributors": [
     "Blum, Brian <brianbl@microsoft.com>"
   ],
-  "version": "1.0.0-preview",
+  "version": "1.0.1-preview",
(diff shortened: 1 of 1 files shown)


lib/services/mediaServicesManagement/package.json
```diff
@@ -14,8 +14,8 @@
     "node",
     "azure"
   ],
-  "main": "./lib/mediaservicesManagementClient.js",
```
amarzavery: Thanks for the contribution. Can you also update the version to "1.0.1-preview" in this PR?

akshatcm: Done @amarzavery

amarzavery:
@akshatcm - The package has been published to npm.

Issue #2225, opened 2017-07-20
Make ms-rest-azure accept the credentials file generated by az cli


amarzavery:

lostintangent:
This will be amazing :)

amarzavery:
Please read this [comment from Yugang](https://github.com/Azure/azure-xplat-cli/issues/3200#issuecomment-316126109) to get an overview about MSI. The work that is being done in az-cli will be done for other sdks as well.

PR #2231, opened 2017-07-22
support for authentication using sp from auth file

amarzavery:
- @balajikris @yugangw-msft - Please review

### DIFF
--- a/runtime/ms-rest-azure/lib/login.js
+++ b/runtime/ms-rest-azure/lib/login.js
@@ -5,11 +5,13 @@
 
 const adal = require('adal-node');
 const async = require('async');
+const fs = require('fs');
+const msRest = require('ms-rest');
 const azureConstants = require('./constants');
+const AzureEnvironment = require('./azureEnvironment');
 const ApplicationTokenCredentials = require('./credentials/applicationTokenCredentials');
 const DeviceTokenCredentials = require('./credentials/deviceTokenCredentials');
 const UserTokenCredentials = require('./credentials/userTokenCredentials');
-const AzureEnvironment = require('./azureEnvironment');
 const SubscriptionClient = require('./subscriptionManagement/subscriptionClient');
 
 // It will create a DeviceTokenCredentials object by default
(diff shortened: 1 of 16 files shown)


amarzavery:
https://github.com/Azure/azure-sdk-for-node/issues/2225

runtime/ms-rest-azure/package.json
```diff
@@ -22,11 +34,12 @@
     "@types/node": "^7.0.10"
```
amarzavery: min dependency for ms-rest should go to ^2.2.2 as the `_withAuthFile()` function in login.js depends on a function homeDir() added in msRest in this PR. After the CI for first commit is green will update this dep in second commit. until the ms-rest version is not published this PR will fail in the second commit due to this reason.

runtime/ms-rest-azure/test/rpRegistrationFilterTests.js
```diff
@@ -1,19 +1,18 @@
...
-      let rpName = RPRegistrationFilter.checkRPNotRegisteredError(msg);
+      var msg = "{\"error\":{\"code\":\"MissingSubscriptionRegistration\",\"message\":\"The subscription is not registered to use namespace 'Microsoft.Devices'. See https://aka.ms/rps-not-found for how to register subscriptions.\"}}";
+      var rpName = RPRegistrationFilter.checkRPNotRegisteredError(msg);
       rpName.should.equal('Microsoft.Devices');
       done();
     });
     it('should correctly parse the error message from service for an existing subscription where the RP was manually unregistered', (done) => {
-      let msg = "{\"error\":{\"code\":\"MissingSubscriptionRegistration\",\"message\":\"The subscription registration is in 'Unregistered' state. The subscription must be registered to use namespace 'Microsoft.Devices'. See https://aka.ms/rps-not-found for how to register subscriptions.\"}}";
+      var msg = "{\"error\":{\"code\":\"MissingSubscriptionRegistration\",\"message\":\"The subscription registration is in 'Unregistered' state. The subscription must be registered to use namespace 'Microsoft.Devices'. See https://aka.ms/rps-not-found for how to register subscriptions.\"}}";
```
balajikris: is there a reason to go from `let` to `var` ?

runtime/ms-rest-azure/test/credTests.js
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+var assert = require('assert');
```
balajikris: `const` ?

runtime/ms-rest-azure/test/credTests.js
```diff
@@ -0,0 +1,110 @@
...
+
+describe('Login', () => {
+  before((done) => {
+    try {
+      credsObj = JSON.parse(fs.readFileSync(authFilePathUsProd, { encoding: 'utf8' }));
+      customCredsObj = JSON.parse(fs.readFileSync(authFilePathCustom, { encoding: 'utf8' }));
+      var dummyCreds = {
+      tokenAudience: undefined,
+      environment: {},
+      authorizattionScheme: 'Bearer',
```
balajikris: typo. `authorizationScheme`

runtime/ms-rest-azure/test/credTests.js
```diff
@@ -0,0 +1,110 @@
...
+        should.not.exist(err);
+        should.exist(creds);
+        should.exist(subscriptions);
+        should.not.exist(process.env['AZURE_SUBSCRIPTION_ID']);
+        process.env['SUB_ID'].should.equal(credsObj.subscriptionId);
+        done();
+      });
+    });
+
+    it('should add custom env to AzureEnvironment and authenticate against it', (done) => {
```
balajikris: 2 more scenarios come to mind - 
1. authfile with insufficient data (if az-cli encounters a bug while generating auth file)
1. foreign cloud environments? not sure if this is testable with stubs, so i'll leave it to you to decide.

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+  if(filePath.startsWith('~')) {
+    filePath = msRest.homeDir(filePath.slice(1));
+  }
+
+  let content = null, credsObj = {}, optionsForSpSecret = {};
+  try {
+    content = fs.readFileSync(filePath, {encoding: 'utf8'});
+    credsObj = JSON.parse(content);
+    _validateAuthFileContent(credsObj, filePath);
+    if (!process.env[subscriptionEnvVariableName]) {
```
balajikris: move this to outside of the `try-catch`. The code in its current shape reads like - read and parse a json.. if an environment variable isn't set, then set it and the flow breaks there. which makes me wonder, why read and parse at all if the env var is already set. The answer is few lines below here, where we continue to use `credsObj`. So, move the setting of `subscription` to outside the `try-catch` and it doesn't need a try-catch anyway.

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+    let environmentObj = AzureEnvironment[env];
+    if (environmentObj && 
+      environmentObj.managementEndpointUrl && 
+      _foundManagementEndpointUrl(credsObj.managementEndpointUrl, environmentObj.managementEndpointUrl)) {
+      envFound.state = true;
+      envFound.name = environmentObj.name;
+      break;
+    }
+  }
+  if (envFound.state) {
```
balajikris: dont think you need `state` field. can simply check if name is not null or empty?

runtime/ms-rest-azure/test/rpRegistrationFilterTests.js
amarzavery: jshint was complaining. So changed it.

runtime/ms-rest-azure/test/credTests.js
amarzavery: should be ok for now

runtime/ms-rest-azure/test/credTests.js
amarzavery: thanks for pointing

runtime/ms-rest-azure/test/credTests.js
amarzavery: for 2. what i intend to test here is that: if someone wants to test against dogfood or other env that is not predefined then our code can correctly add that environment and try to authenticate using the sp info

amarzavery: for 1. I have the [_validateAuthFileContent](https://github.com/Azure/azure-sdk-for-node/pull/2231/files#diff-435cfc34b8c2083d4a454c469df87d66R473) that throws an error if something is incorrect.

I can add a test for that scenario.

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+
+  let filePath = options.filePath;
+  let authLocation = process.env[azureConstants.AZURE_AUTH_LOCATION];
+  let subscriptionEnvVariableName = options.subscriptionEnvVariableName || 'AZURE_SUBSCRIPTION_ID';
+  if (!filePath) {
+    if (!authLocation) {
+      let msg = `Either provide an absolute file path to the auth file or set/export the environment variable - ${azureConstants.AZURE_AUTH_LOCATION}.`;
+      return callback(new Error(msg));
+    }
+    filePath = authLocation;
```
amarzavery: if the filePath is not provided then we look for the authLocation (env. var). If that is not provided as well then we throw. If it is provided then we set the value of authlocation env. var to the filePath parameter.

runtime/ms-rest-azure/lib/login.js
amarzavery: The environment variable that we are setting over here is the env. variable for subscriptionId. It is not the env. var for auth location. Node sdk has a recommended set of environment variables for things like subscriptionId, clientId, username, password, domain, etc.
So in this case, if the user provides his custom environment variable name that is being used in his app then we use that env. variable to set the value of subscriptionId. IF not then we set the subscriptionId to AZURE_SUBSCRIPTION_ID env. variable.
This subscription Id will be the only subscriptionId that the customer can use while instantiating the client since the SP is linked to that subscription. Using any other subscriptionId will be setting him up for failure.

I know this is a bit confusing. But whatever I said above, does that make sense or clear the initial question that you asked?

amarzavery: I think the check is incorrect. It should actually be like this:

```js
//just set the environment variable to correct subscriptionId from the auth file if it is not present.
if (process.env[subscriptionEnvVariableName] && process.env[subscriptionEnvVariableName] !== credsObj.subscriptionId) {
  process.env[subscriptionEnvVariableName] = credsObj.subscriptionId;
}
```

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+  if (!callback && typeof options === 'function') {
+    callback = options;
+    options = {};
+  }
+  if (!options) options = { filePath: '' };
+  if (!callback) {
+    throw new Error('callback cannot be null or undefined.');
+  }
+
+  let filePath = options.filePath;
```
yugangw-msft: in case it helps, will the following code simplify it?
```javascript
    let filePath = options.filePath || process.env[azureConstants.AZURE_AUTH_LOCATION]` 
    if !filepath:
        return callback(new Error(...))
```

runtime/ms-rest-azure/lib/login.js
yugangw-msft: Hope I understand the context correctly, the subscription from the auth file should be precedent the environment variable. You can overwrite the environment variable here, but that should be just an implementation detail being transparent to users.  Same with other fields. 
If I got it right, then we don't need an `if` condition, just overwrite it, optionally emit out a warning

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+        } else {
+          envParams[key] = credsObj[key];
+        }
+      }
+    }
+    if (!envParams.activeDirectoryResourceId) {
+      envParams.activeDirectoryResourceId = credsObj.managementEndpointUrl;
+    }
+    if (!envParams.portalUrl) {
+      envParams.portalUrl = 'https://portal.azure.com';
```
yugangw-msft: This is not an environment aware url. If SDK never uses it, which I believe is the case, then maybe we just remove it from the environment

runtime/ms-rest-azure/test/data/auth-customEnv.json
```diff
@@ -0,0 +1,12 @@
+{
+  "clientId": "0b868e09-1547-4721-9a57-705a15328409",
+  "clientSecret": "005937fd-3a0f-4cf2-84da-da7388241b9f",
```
yugangw-msft: Assume this is a faked secret?

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+  if (!credsObj.resourceManagerEndpointUrl) {
+    throw new Error(`"resourceManagerEndpointUrl" is missing from the auth file: ${filePath}.`);
+  }
+  if (!credsObj.activeDirectoryGraphResourceId) {
+    throw new Error(`"activeDirectoryGraphResourceId" is missing from the auth file: ${filePath}.`);
+  }
+  if (!credsObj.sqlManagementEndpointUrl) {
+    throw new Error(`"sqlManagementEndpointUrl" is missing from the auth file: ${filePath}.`);
+  }
+  if (!credsObj.galleryEndpointUrl) {
```
amarzavery: ok

runtime/ms-rest-azure/lib/login.js
```diff
@@ -468,4 +470,232 @@ exports.withServicePrincipalSecretWithAuthResponse = function withServicePrincip
...
+    throw new Error(`"galleryEndpointUrl" is missing from the auth file: ${filePath}.`);
+  }
+  if (!credsObj.managementEndpointUrl) {
```
amarzavery: ok

runtime/ms-rest-azure/test/data/auth-customEnv.json
amarzavery: yes

amarzavery:
#2245 and #2247

PR #2236, opened 2017-07-27
Updating content per guidance

dend:


fiveisprime:
Some context: the ARM API for intune is no longer relevant and the SDK module should be deprecated. This PR removes the documentation for the intune module from the reference pages on the docs site.

Issue #2239, opened 2017-08-01
no componentVersion of HDInsight cluster


voidfunction:
Based on https://github.com/Azure/azure-rest-api-specs/blob/current/specification/hdinsight/resource-manager/Microsoft.HDInsight/2015-03-01-preview/cluster.json#L398, `componentVersion`  is a field of rest API but we can acquire it from node SDK.

PR #2240, opened 2017-08-01
componentVerions support for HDInsight cluster

voidfunction:
fix: #2239

### DIFF
--- a/lib/services/HDInsight2/lib/hDInsightManagementClient.js
+++ b/lib/services/HDInsight2/lib/hDInsightManagementClient.js
@@ -4094,7 +4094,17 @@ var ClusterOperations = ( /** @lends ClusterOperations */ function() {
                     var kindInstance = kindValue;
                     clusterDefinitionInstance.clusterType = kindInstance;
                   }
-                  
+
+                  var componentVerionsInDefinition = clusterDefinitionValue['componentVersion'];
+                  if (componentVerionsInDefinition !== null && componentVerionsInDefinition !== undefined) {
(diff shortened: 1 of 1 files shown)


lib/services/HDInsight2/lib/hDInsightManagementClient.js
```diff
@@ -4094,7 +4094,17 @@ var ClusterOperations = ( /** @lends ClusterOperations */ function() {
...
+                  var componentVerionsInDefinition = clusterDefinitionValue['componentVersion'];
+                  if (componentVerionsInDefinition !== null && componentVerionsInDefinition !== undefined) {
+                      var componentVerionsInstance = {};
+                      for (var item in componentVerionsInDefinition) {
+                        componentVerionsInstance[item] = componentVerionsInDefinition[item];
+                      }
+
+                      clusterDefinitionInstance.componentVerions = componentVerionsInstance;
+                  }
+
```
amarzavery: @voidfunction - This library is autogenerated. Moreover this is autogenerated from the old hydra-specs using the old codegenerator hydra.
The spec that you are referring to is the new swagger spec. This library needs to be autogenerated using the new code generator Autorest.

voidfunction: Thanks. So what should I do is just waiting for the update?

amarzavery: @voidfunction  Are you from the hdInsight team ? If so then please contact Shefali Vohra regarding this.

PR #2249, opened 2017-08-18
MSI Authentication (scope: VMs)

balajikris:
Basing this PR on top of Amar's AuthFromFile PR which is still open. 

Implementation of this PR is in the following commits: 
1. b1c952d -- implement MSI Token Credentials and add login methods
1. 0cb92a0 -- tests

### DIFF
--- a/runtime/ms-rest-azure/lib/credentials/msiTokenCredentials.js
+++ b/runtime/ms-rest-azure/lib/credentials/msiTokenCredentials.js
@@ -1,0 +1,98 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information. 
+
+'use strict';
+const request = require('request');
+const Constants = require('ms-rest').Constants;
+
+class MSITokenCredentials {
+  constructor(domain, options) {
+
+    if (!Boolean(domain) || typeof domain.valueOf() !== 'string') {
+      throw new Error('domain must be a non empty string.');
+    }
+
+    if (!options) {
+      options = {};
+    }
+
+    if (!options.port) {
+      options.port = 50342; // default port where token service runs.
+    } else if (typeof options.port.valueOf() !== 'number') {
+      throw new Error('port must be a number.');
+    }
+
+    if (!options.resource) {
+      options.resource = "https://management.azure.com";
+    } else if (typeof options.resource.valueOf() !== 'string') {
+      throw new Error('resource must be a uri.');
+    }
+
+    this.domain = domain;
+    this.port = options.port;
+    this.resource = options.resource;
+    this.aad_endpoint = "https://login.microsoftonline.com";
+  }
+
+  /**
+   * Tries to get the token from cache initially. If that is unsuccessfull then it tries to get the token from ADAL.
+   * @param  {function} callback  The callback in the form (err, result)
+   * @return {function} callback
+   *                       {Error} [err]  The error if any
+   *                       {object} [tokenResponse] The tokenResponse (token_type and access_token are the two important properties). 
+   */
+  getToken(callback) {
+    const postUrl = `http://localhost:${this.port}/oauth2/token`;
+    const reqOptions = this.prepareRequestOptions();
+    request.post(postUrl, reqOptions, (err, response, body) => {
+      if (err) {
+        return callback(err);
+      }
+      try {
+        let tokenResponse = JSON.parse(body);
+        if (!tokenResponse.token_type) {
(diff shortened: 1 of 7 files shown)


lostintangent:
This looks awesome! In what environments can this new login method be used? Just VMs that we're provisioned with a managed identity?

balajikris:
Hey @lostintangent -- thanks! Currently, we're looking to bring this on for VMs that have been provisioned with the MSI extension. @yugangw-msft is my best guess as to someone who _may_ have a better idea about how or when we'll onboard other services with MSI ?

package.json
```diff
@@ -1,157 +1,157 @@
...
+    "grunt": "~0.4",
+    "grunt-contrib-connect": "^0.10.1",
+    "grunt-contrib-symlink": "^0.3.0",
+    "grunt-devserver": "~0.6",
+    "grunt-jsdoc": "~0.6",
+    "grunt-gh-pages": "1.1.0",
+    "jshint": "2.9.4",
+    "minami": "git://github.com/devigned/minami#master",
+    "mocha": "^3.2.0",
+    "nock": "^9.0.14",
```
amarzavery: dont change the nock version for rollup azure.
Instead add  nock as a dev dependency in ms-rest-azure's package.json and this test over there. Since this is more of a unit test and not a scenario/live test anymore.

runtime/ms-rest-azure/lib/credentials/msiTokenCredentials.js
```diff
@@ -12,25 +12,25 @@ class MSITokenCredentials {
       throw new Error('domain must be a non empty string.');
```
amarzavery: Please take a look at the flow implemented for MSI auth flow in CLI. 
- https://github.com/Azure/azure-cli/blob/master/src/azure-cli-core/azure/cli/core/_profile.py#L532-L592.

The CLI asks for the tenantId against which the token is required. We are not asking that from the customer.

runtime/ms-rest-azure/lib/login.js
```diff
@@ -1,699 +1,732 @@
...
+    throw new Error('callback cannot be null or undefined.');
+  }
+  const creds = new MSITokenCredentials(domain, options)
+  creds.getToken(function (err) {
+    if (err) return callback(err);
+    return callback(null, creds);
+  });
+}
+
+exports.withMSI = function withMSI(domain, options, optionalCallback) {
```
amarzavery: missing documentation

amarzavery: export this in lib/msRestAzure.js

amarzavery: update the changelog

runtime/ms-rest-azure/lib/credentials/msiTokenCredentials.js
balajikris: Isn't tenantID also called as Domain? I could have used the term `tenantID` but all the other login methods call it `domain`, so I stuck to it for consistency. Unless I'm completely off here.

runtime/ms-rest-azure/lib/login.js
balajikris: work in progress.

runtime/ms-rest-azure/test/msiTokenCredentialTests.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+// all the requires
+var should = require('should');
+var nock = require('nock');
+var MSITokenCredential = require('../lib/credentials/msiTokenCredentials');
+
+
+describe('MSI Authentication', function () {
+    before(function (done) {
```
amarzavery: indentation 2 spaces

runtime/ms-rest-azure/package.json
```diff
@@ -28,7 +28,7 @@
     "async": "0.2.7",
     "uuid": "^3.0.1",
     "adal-node": "^0.1.17",
-    "ms-rest": "^2.2.2",
+    "ms-rest": "^2.2.1",
```
amarzavery: why was this changed to 2.2.1? We intend to publish this with new version as the min dependency.

balajikris: good catch -- had that to get tests passing, verified that CI passes (https://travis-ci.org/Azure/azure-sdk-for-node/jobs/270179285) .. Will revert this now.

balajikris:
Status: All tests pass w/ the right version of ms-rest. See [here](https://travis-ci.org/Azure/azure-sdk-for-node/builds/270179284?utm_source=github_status&utm_medium=notification).

PR #2254, opened 2017-08-28
Updating Notification Hubs SDK to include support for APNS Headers

aimankhan:


lib/services/serviceBus/lib/apnsservice.js
```diff
@@ -39,7 +39,7 @@ var registrationResult = require('./models/registrationresult');
 * @param {NotificationHubService} notificationHubService The notification hub service.
 */
 function ApnsService(notificationHubService) {
-  this.notificationHubService = notificationHubService;
+    this.notificationHubService = notificationHubService;
```
amarzavery: tabbing is 2 spaces in javascript

PR #2257, opened 2017-08-29
Update Notification Hubs SDK

aimankhan:


aimankhan:
@amarzavery - it looks like there are some KeyVault test failures... I can't see how my changes would impact the KeyVault's SDK. Any ideas on if this is expected?

PR #2268, opened 2017-09-21
Adding OAuth2ImplicitFlow property to create and update Application

ppathan:
Adding OAuth2ImplicitFlow property to create and update Application


amarzavery:
Thanks for the PR.  As I said the sdks in this repo are autogenerated. This property needs to be added over here: https://github.com/Azure/azure-rest-api-specs/blob/current/specification/graphrbac/data-plane/1.6/graphrbac.json#L1842. Once that change is merged, then we can autogenerate the sdk and get the change in.

amarzavery:
Please send a PR with the property addition to the azure-rest-api-specs repo.

PR #2270, opened 2017-09-23
updating graph library with the latest swagger changes as of 9/22/2017

amarzavery:


amarzavery:
@ppathan - This is the new autogenerated version of the graph library.

PR #2279, opened 2017-09-30
removing mocha html-cov and replacing it with istanbulJS

cpenarrieta:
mochaJs have [removed](https://github.com/mochajs/mocha/issues/2356) their coverage implementation. This is why after running `npm run coverage` we get the following error `Error: invalid reporter "html-cov"`.
I'm adding [istanbulJs](https://istanbul.js.org/) in order to fix the coverage in this project.
You can test this PR by running `npm run coverage`, you should see the following result:
image
A detailed html report will be created in `/coverage/index.html`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ test/tmp/*
 .nuget
 packages
 
+# Coverage report
+coverage
+.nyc_output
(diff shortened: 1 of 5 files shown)


amarzavery:
LGTM

PR #2281, opened 2017-10-02
Batch sdk test updates

annatisch:


amarzavery:
@annatisch - I was not able to see the package version update in `lib/services/batch/package.json`. Can you please update that as well?

PR #2299, opened 2017-10-19
Updated activeDirectoryEndpointUrl for AzureUSGovernment environment …

bernie-msft:
…and updated the change log


bernie-msft:
#sign-off

Issue #2301, opened 2017-10-23
linkedServices.createOrUpdate() what is the format of the linkedService parameter?


kitsonswannff:
Hi There,

I'm trying to create a Data Factory with an Amazon Redshift linked service. I can successfully create the data factory, but can't get the linked services working. 

I'm calling the `dataFactoryManagement.linkedServices.createOrUpdate()` method, but I cannot seem to get the right format for the `linkedService` parameter. It says it's an object. I've looked at the REST api documentation which seems to indicate this is the format:

```
{
    "name": "AmazonRedshiftLinkedService",
    "properties":
    {
        "type": "AmazonRedshift",
        "typeProperties":
        {
            "server": "< The IP address or host name of the Amazon Redshift server >",
            "port": <The number of the TCP port that the Amazon Redshift server uses to listen for client connections.>,
            "database": "<The database name of the Amazon Redshift database>",
            "username": "<username>",
            "password": "<password>"
        }
    }
}
```

I've created a function that calls this method like so

```
function createRedshiftLinkedService(callback) {

  var linkedServiceName = "SourceRedshiftLinkedService";

  var redshiftLinkedServiceProperties =  {
      properties: {
        type: "AmazonRedshift",
        typeProperties: {
            server: redshiftServer,
            port: redshiftServerPort,
            database: redshiftDatabase,
            username: redshiftUserName,
            password: redshiftPassword
        }
    }
  };

  console.log(redshiftLinkedServiceProperties);

  return dfClient.linkedServices.createOrUpdate(resourceGroupName, dataFactoryName, linkedServiceName, redshiftLinkedServiceProperties, callback);

}
```

When it runs I get the following error:

`Error: Error "server" cannot be null or undefined in "linkedService.properties".`

I've tried changing the format and moving the `server` variable to the next level up. I then complains that there's no `server` variable under `typeProperties`. If I put it in both places, it says `typeProperties.server should be an object`

Can you give me any guidance on the proper format to provide the linkedService parameter? Ideally, could you give a minimum example of actually creating a data factory using the node.js sdk, including the linked services. The only example in the docs just lists the data factories, which isn't helpful.

Thanks.

amarzavery:
Hi @kitsonswannff - Thanks for reporting the issue. There is a bug in serialization in the sdk.

[...]

PR #2302, opened 2017-10-25
relax type validation for object types

amarzavery:


amarzavery:
fixes the issue in #2301

PR #2328, opened 2017-11-14
Add innerError field to CloudError contract

demyanenko:
ARM error schema conforms to the [OData 4.0](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091) spec, which declares the innerError field as an extension point.

I'm working on Azure Policy, and there is an upcoming change to ARM after which any PUT/PATCH request to any resource provider may return an error containing the innerError field in case the request is violating one of the policies. Because it's a part of ARM contract and is not specific to an RP, CloudError contract should be the right place for this change.

### DIFF
--- a/runtime/ms-rest-azure/lib/cloudError.js
+++ b/runtime/ms-rest-azure/lib/cloudError.js
@@ -14,6 +14,8 @@
  * @member {string} [target] The target of the error
  * 
  * @member {array} [details] An array of CloudError objects specifying the details
+ * 
+ * @member {Object} [innerError] The inner error parsed from the body of the http error response
  */
 class CloudError extends Error {
   constructor(parameters) {
@@ -42,6 +44,10 @@ class CloudError extends Error {
           tempDetails.push(element);
         });
         this.details = tempDetails;
+      }
+
+      if (parameters.innerError) {
+        this.innerError = JSON.parse(JSON.stringify(parameters.innerError));
       }
     }
   }
@@ -95,11 +101,19 @@ class CloudError extends Error {
                 }
               }
             }
+          },
+          innerError: {
+            required: false,
+            serializedName: 'innerError',
+            type: {
+              name: 'Object'
+            }
           }
         }
       }
     };
   }
+
   /**
    * Serialize the instance to CloudError schema
    *
@@ -129,6 +143,10 @@ class CloudError extends Error {
         deserializedArray.push(element1);
       });
       payload.error.details = deserializedArray;
+    }
+
+    if (this.innerError) {
+      payload.error.innerError = JSON.parse(JSON.stringify(this.innerError));
     }
     return payload;
   }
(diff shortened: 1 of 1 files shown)


runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -95,11 +101,19 @@ class CloudError extends Error {
                 }
               }
             }
+          },
+          innerError: {
+            required: false,
+            serializedName: 'innerError',
```
amarzavery: The [odata v4.0 spec](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091) and the [Azure Resource Manager RPC spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/common-api-details.md#error-response-content) says that the name of the property is `innererror`.

> The value for the innererror name/value pair MUST be an object. The contents of this object are service-defined. Usually this object contains information that will help debug the service. The innererror name/value pair SHOULD only be used in development environments in order to guard against potential security concerns around information disclosure.

The PR states the name of the property is innerError (camelCase).

runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -43,6 +45,10 @@ class CloudError extends Error {
         });
         this.details = tempDetails;
       }
+
+      if (parameters.innerError) {
+        this.innerError = JSON.parse(JSON.stringify(parameters.innerError));
```
amarzavery: What is the need to stringify the object and then parse it again if it is already a JSON object. I understand that this is usually done to clone the object. However, I dont understand the need to do this in the current scenario.
Can you please explain?

runtime/ms-rest-azure/lib/cloudError.js
demyanenko: You are right. Changed the name to innererror

runtime/ms-rest-azure/lib/cloudError.js
demyanenko: The details field is being reconstructed in every function. I assumed that we can assign other fields because the strings are immutable, and the complex objects should be deep cloned. I'm not too familiar with the code, so please let me know if the deep cloning is unnecessary.

amarzavery: No worries. It is ok to keep it like that. Anyways serialize() and deserialize() functions in CloudError are for backwards compatibility. Newer versions of the generated sdk do not use them anymore. They only depend on the mapper and then serialize/deserialize accordingly.
So you are good for now.

Can you also update the Changelog.md (https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest-azure/Changelog.md)
 - Added `innererror` property to `CloudError`

PR #2329, opened 2017-11-15
Updated typings to include needed properties that already exist

christav:
Adding properties to the typings in a couple of types to expose essential properties. For example, every one of the credentials classes in this repos assigned directly to webResource.headers, and without response.statusCode HttpOperationResponse is basically useless.

### DIFF
--- a/runtime/ms-rest/index.d.ts
+++ b/runtime/ms-rest/index.d.ts
@@ -3,7 +3,7 @@ import * as http from 'http';
 
 /**
  * REST request options
- *  
+ *
  * @property {object.<string, string>} customHeaders - Any additional HTTP headers to be added to the request
  * @proprerty {boolean} [jar] - If true, remember cookies for future use
  */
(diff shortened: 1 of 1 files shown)


runtime/ms-rest/index.d.ts
```diff
@@ -16,23 +16,40 @@ export interface ClientRequestOptions extends RequestOptions {
...
+ * HttpResponse wrapper that encapsulates the
+ * response stream plus available status codes/messages
+ *
+ * @property {Object} headers       - response headers returned from the server, duplicate filtered
+ *                                    as described in https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_message_headers
+ * @property {Array} [rawHeaders]   - Original unprocessed response headers
+ * @property {Number} statusCode    - HTTP status code for response
+ * @property {String} statusMessage - HTTP status response message
+ */
+export interface Response extends stream.Readable {
```
amarzavery: Rather than defining our own type, it would be better to expose the RequestResponse interface from the request typings.
We have a dependency on `@types/request`.

```typescript
import { RequestResponse } from 'request';

. . . 
export interface HttpOperationResponse<T> {
  request: WebResource;
  response: RequestResponse;
  body: T;
}
 . . .
```

amarzavery:
Please update the definition of response in ServiceCallback as well,

```typescript
export interface ServiceCallback<TResult> { (err: Error | ServiceError, result?: TResult, request?: WebResource, response?: RequestResponse): void }
```

christav:
@amarzavery I updated according to your comments with one difference - I used the http.IncomingMessage instead of RequestResponse as the type. I don't think it's a good idea to couple this library's public API to the Request library, but everything's going to be using http.IncomingMessage under the hood.

amarzavery:
@christav - I do see your point. Tomorrow if we plan to change the underlying request library for some reason then our exposed API would not get affected. 

One question though:
This is the definition of [RequestResponse](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/request/index.d.ts#L183) in the d.ts file of `@types/request`, which also extends from http.IncomingMessage.

```typescript
export interface RequestResponse extends http.IncomingMessage {

		request: ResponseRequest;

		body: any;

		timingStart?: number;

		timings?: {

			socket: number;

			lookup: number;

			connect: number;

			response: number;

			end: number;

		};

		timingPhases?: {

			wait: number;

			dns: number;

			tcp: number;

			firstByte: number;

			download: number;

			total: number;

		};

	}
```

- By not exposing this as the type, only downside would be that someone would not get the typings for the properties defined in `RequestResponse`.  Will that be ok?

christav:
I don't see it as an issue. Other than the timing stuff, the other extended info is already available in other ways through the existing interface. And the timing is very request-specific, so I'd be fine doing a cast there.

amarzavery:
I see, that sounds cool. Once the CI passes, I shall merge this PR.

PR #2343, opened 2017-11-30
Generating client for azure consumption api version 2017-11-30

sandeepnl:


package.json
```diff
@@ -41,7 +41,7 @@
     "azure-arm-cognitiveservices": "2.0.0",
     "azure-arm-commerce": "1.1.0",
     "azure-arm-compute": "4.0.0",
-    "azure-arm-consumption": "1.0.0-preview",
+    "azure-arm-consumption": "2.0.0",
```
amarzavery: @sandeepnl - Please remove this change for now. It is causing the CI to fail (npm install fails because it cannot find this version, which is true because we haven't published it yet.) We will update this when we publish the rollup version. (I will update the documentation to not ask for this change).

PR #2356, opened 2017-12-13
Serialization changes, remove requirement for polymorphic discriminator type in json

williexu:
- follow up from https://github.com/Azure/azure-sdk-for-node/pull/2332, deals with missing polymorphic 
discriminator in response
- during runtime, discriminator property is populated with the expected value if missing.

### DIFF
--- a/runtime/ms-rest/lib/serialization.js
+++ b/runtime/ms-rest/lib/serialization.js
@@ -222,12 +222,16 @@ function serializeCompositeType(mapper, object, objectName) {
           parentObject = parentObject[pathName];
         });
 
-        //make sure required properties of the CompositeType are present
-        if (modelProps[key].required && !modelProps[key].isConstant) {
-          if (object[key] === null || object[key] === undefined) {
+        if (object[key] === null || object[key] === undefined) {
+          if (modelProps[key].isPolymorphicDiscriminator){
+            //add expected polymorphic discriminator when serializing if it is missing
+            parentObject[propName] = mapper.serializedName;
+          } else if (modelProps[key].required && !modelProps[key].isConstant) {
+            //required properties of the CompositeType must be present
             throw new Error(`${key}" cannot be null or undefined in "${objectName}".`);
           }
         }
+
         //make sure that readOnly properties are not sent on the wire
         if (modelProps[key].readOnly) {
           continue;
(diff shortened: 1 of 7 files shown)


amarzavery:
add tests for this scenario in ms-rest.

runtime/ms-rest/lib/serialization.js
```diff
@@ -580,9 +576,6 @@ function _getPolymorphicMapperStringVersion(mapper, object, objectName) {
       throw new Error(`${objectName}" cannot be null or undefined. "${mapper.type.polymorphicDiscriminator}" is the ` +
         `polmorphicDiscriminator and is a required property.`);
     }
-    if (object[mapper.type.polymorphicDiscriminator] === null || object[mapper.type.polymorphicDiscriminator] === undefined) {
-      throw new Error(`No discriminator field "${mapper.type.polymorphicDiscriminator}" was found in "${objectName}".`);
```
amarzavery: same as above.

runtime/ms-rest/lib/serialization.js
```diff
@@ -554,10 +554,6 @@ function _getPolymorphicMapperObjectVersion(mapper, object, objectName, mode) {
         `"${mapper.type.polymorphicDiscriminator[polymorphicPropertyName]}" is the ` +
         `polmorphicDiscriminator and is a required property.`);
     }
-    if (object[mapper.type.polymorphicDiscriminator[polymorphicPropertyName]] === null ||
-      object[mapper.type.polymorphicDiscriminator[polymorphicPropertyName]] === undefined) {
-      throw new Error(`No discriminator field "${mapper.type.polymorphicDiscriminator[polymorphicPropertyName]}" was found in "${objectName}".`);
```
amarzavery: I would recommend not removing this check. Rather do the following:
```javascript
if (object[mapper.type.polymorphicDiscriminator[polymorphicPropertyName]] === null ||
  object[mapper.type.polymorphicDiscriminator[polymorphicPropertyName]] === undefined) {
  return mapper;
}
```
The reason being few lines down on line 561 
```javascript
indexDiscriminator = mapper.type.uberParent + '.' + object[mapper.type.polymorphicDiscriminator[polymorphicPropertyName]];
```
`indexDiscriminator` will be `<the value of the parent model>.undefined` and that would be weird.

We would eventually return the same mapper. So why not save the execution time and return the mapper early.

Does that make sense?

amarzavery:
also rebase with upstream master. That should help get the changes in #2357 that just got merged.

amarzavery:
@williexu - Please add a test for this in the nodejs code generator as well.

runtime/ms-rest/lib/serialization.js
```diff
@@ -222,12 +222,16 @@ function serializeCompositeType(mapper, object, objectName) {
           parentObject = parentObject[pathName];
         });
 
-        //make sure required properties of the CompositeType are present
-        if (modelProps[key].required && !modelProps[key].isConstant) {
-          if (object[key] === null || object[key] === undefined) {
+        if (object[key] === null || object[key] === undefined) {
+          if (modelProps[key].isPolymorphicDiscriminator){
+            //add expected polymorphic discriminator when serializing if it is missing
+            parentObject[propName] = mapper.serializedName;
```
amarzavery: object[key] = mapper.serializedName;

runtime/ms-rest/lib/serialization.js
```diff
@@ -507,13 +511,16 @@ function deserializeCompositeType(mapper, responseBody, objectName) {
...
         //paging
-        if (Array.isArray(responseBody[key]) && modelProps[key].serializedName === '') {
-          propertyInstance = responseBody[key];
+        if (Array.isArray(responseValue) && modelProps[key].serializedName === '') {
+          propertyInstance = responseValue;
           instance = exports.deserialize.call(this, propertyMapper, propertyInstance, propertyObjectName);
         } else if ((propertyInstance !== null && propertyInstance !== undefined) || (propertyMapper && propertyMapper.isConstant)) {
           serializedValue = exports.deserialize.call(this, propertyMapper, propertyInstance, propertyObjectName);
           instance[key] = serializedValue;
+        } else if (!responseValue && propertyMapper.isPolymorphicDiscriminator){
```
amarzavery: same as above. Let exports.deserialize.call() do its job. set responeBody[key] to mapper.serializedName around line 513.

amarzavery:
Please update the patch version in package.json and the changelog.

PR #2371, opened 2017-12-22
Fix to address different underscore_cased properties in msi tokenResponses

williexu:


williexu:
This will be resolved by camelCasing the properties in the runtime after the response is received:

PR #2405, opened 2018-01-31
Introduced new operations and updated existing UsageDetails operation

shbha:
1) Introduced new models and operations related to 3 API operations:
Budgets,
Marketplace, and
PriceSheet

2) Updated UsageDetails - changed scope parameter to accept separate parameters - subscriptionId and BillingPeriod


lib/services/consumptionManagement/lib/consumptionManagementClient.d.ts
```diff
@@ -40,14 +44,18 @@ declare class ConsumptionManagementClient extends AzureServiceClient {
    * @param {boolean} [options.generateClientRequestId] - When set to true a unique x-ms-client-request-id value is generated and included in each request. Default is true.
    *
    */
-  constructor(credentials: ServiceClientCredentials, subscriptionId: string, baseUri?: string, options?: AzureServiceClientOptions);
+  constructor(credentials: ServiceClientCredentials, subscriptionId: string, resourceGroupName: string, budgetName: string, baseUri?: string, options?: AzureServiceClientOptions);
```
amarzavery: @shbha Why are resourceGroupName and budgetName properties of the client? Shouldn't they be method parameters? This will break existing clients.

amarzavery:
@shbha - Please update the package version in [this file](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/consumptionManagement/package.json).

lib/services/consumptionManagement/lib/consumptionManagementClient.d.ts
amarzavery: @shbha - The consumption api spec has been updated in this PR  https://github.com/Azure/azure-rest-api-specs/pull/2387. Please regenerate the sdk.

lib/services/consumptionManagement/lib/consumptionManagementClient.js
```diff
@@ -1,8 +1,4 @@
 /*
- * Copyright (c) Microsoft Corporation. All rights reserved.
- * Licensed under the MIT License. See License.txt in the project root for
- * license information.
- *
```
amarzavery: if you are using autorest to generate the client manually then please use
```
--license-header=MICROSOFT_MIT_NO_VERSION
```

also make sure to use `--azure-arm`

amarzavery:
make sure to bump the major version of the version in the package.json file https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/consumptionManagement/package.json

lib/services/consumptionManagement/lib/consumptionManagementClient.js
shbha: just noticed your message. I'll make the change now.

ms-premp:
@amarzavery we will have to regenerate NODE SDK due to changes in swagger. At present we are waiting on swagger PR https://github.com/Azure/azure-rest-api-specs/pull/2394 to complete. Appreciate if you can help on reviewing that and then we can regenerate the SDK. Thanks in advance.

shbha:
@amarzavery - The PR has the latest changes related to the swagger-fix I mentioned to you earlier. The package version has been updated to 3.0.0. Thanks.

PR #2432, opened 2018-02-26
Apiversion

amolr:
Increase the API version for azure-common so that the new api version is appended to the Notification Hub request


amarzavery:
Did you make any changes to azure-common? We do not release a new version of the rollup package frequently. If this is being used from `azure-sb` or any other package for notification hubs then we would expect you to update the dependency version of azure-common over there.

amarzavery:
@amolr - Please take a look at the comments in this PR as well.

amolr:
Hi Amar,

There is no change to azure-common but thats the latest version that is required. I see that the package.json of azure-sb(services/servicesbus) is configured as follows
 "dependencies": {
    "azure-common": "^0.9.14",
    "underscore": "^1.8.3",
    "wns": "~0.5.3",
    "mpns": "2.1.3"
  }

But I still dont see azure-common 0.9.19 getting downloaded. This is version we require to be downloaded. Thats the reason why I have updated the package.json at the root level. Also, when customers download the nuget package for Azure node they should be able to get the latest version of azure-common right. NotificationHubs requires certain version of azure-common so that all its requests are successful.
Can you let us know if we are missing something else

amarzavery:
@amolr - That is exactly what I am saying. 

Right now the minimum version is set to 0.9.14. Thus if npm for some reason gives you 0.9.14 it is not wrong. Hence, when you update the dependency of azure-common with "^0.9.19" in the package "azure-sb" it will make sure that minimum version is 0.9.19 or greater than that.

It is good to update the version of azure-common in the root package.json. However, we do not frequently ship the rollup package. If you update the dependency  version of azure-common in azure-sb then folks who directly install `"azure-sb"`will get a version of azure-common that is greater than 0.9.19

I hope I have cleared the confusion.

Tip: When you want to test the installation version, please make sure you delete the node_modules folder. Also make sure you have at least 8.9.4 version of node.js.

amarzavery:
I am asking you to update [this file](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/serviceBus/package.json#L27) `"azure-common": "^0.9.19"`

amarzavery:
@amolr - I am closing this PR. Make all the changes in the other PR. It is easy to capture the changes.

PR #2436, opened 2018-02-27
Adding support for tags update registration

amolr:
Adding support to update registration tags

### DIFF
--- a/lib/services/serviceBus/lib/notificationhubservice.js
+++ b/lib/services/serviceBus/lib/notificationhubservice.js
@@ -290,7 +290,7 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
       delete registration.Expiry;
   }
 
-  if (registration.RegistrationId !== null || registration.RegistrationId !== undefined) {
+  if (registration) {
       delete registration.RegistrationId;
     }
 
@@ -343,7 +343,7 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
       delete registration.Expiry;
   }
 
-  if (registration.RegistrationId !== null || registration.RegistrationId !== undefined) {
+  if (registration) {
       delete registration.RegistrationId;
   }
 
@@ -355,7 +355,7 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
         newRegistration[key] = registration[key];
     }
 
-    if(registration.BodyTemplate !== null || registration.BodyTemplate !== undefined)
+    if(registration.BodyTemplate !== null && registration.BodyTemplate !== undefined)
     newRegistration.BodyTemplate = '<![CDATA[' + registration.BodyTemplate + ']]>';
 
     var registrationXml = registrationResult.serialize(registrationType, newRegistration);
--- a/lib/services/serviceBus/lib/notificationhubservice.js
+++ b/lib/services/serviceBus/lib/notificationhubservice.js
@@ -282,16 +282,26 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
 
   var webResource = WebResource.put(this.hubName + '/registrations/' + registration.RegistrationId);
 
-  registration = _.clone(registration);
   var registrationType = registration[Constants.ATOM_METADATA_MARKER]['ContentRootElement'];
   delete registration[Constants.ATOM_METADATA_MARKER];
   delete registration.ExpirationTime;
 
   if (!registration.Expiry) {
-    delete registration.Expiry;
+      delete registration.Expiry;
   }
 
-  var registrationXml = registrationResult.serialize(registrationType, registration);
+  if (registration.RegistrationId != null) {
+      delete registration.RegistrationId;
+    }
+
+    var newRegistration = {};
+    newRegistration.Tags = registration.Tags;
+    for (var key in registration) {
+        if (key != "Tags")
+            newRegistration[key] = registration[key];
+    }
+
+    var registrationXml = registrationResult.serialize(registrationType, newRegistration);
 
   if (options && options.etag) {
     webResource.withHeader(HeaderConstants.IF_MATCH, options.etag);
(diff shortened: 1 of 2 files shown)


lib/services/serviceBus/package.json
```diff
@@ -11,7 +11,7 @@
     "Tavares, Chris <ctavares@microsoft.com>",
     "Kulshrestha, Ankur <ankurkul@microsoft.com>"
   ],
-  "version": "0.10.5",
+  "version": "0.10.6",
```
amarzavery: Why dont you update the minimum version of azure-common dependency to "^0.9.19" over here as well.

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -282,16 +282,26 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
...
   delete registration[Constants.ATOM_METADATA_MARKER];
   delete registration.ExpirationTime;
 
   if (!registration.Expiry) {
-    delete registration.Expiry;
+      delete registration.Expiry;
   }
 
-  var registrationXml = registrationResult.serialize(registrationType, registration);
+  if (registration.RegistrationId != null) {
```
amarzavery: In javascript use `!==` please check [this link](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons) for better explanation.

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -282,16 +282,26 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
...
+      delete registration.RegistrationId;
+    }
+
+    var newRegistration = {};
+    newRegistration.Tags = registration.Tags;
+    for (var key in registration) {
+        if (key != "Tags")
```
amarzavery: same as above

lib/services/serviceBus/lib/notificationhubservice.js
amarzavery: Wouldn't you want to check for `undefined` as well? In javascript `null` and `undefined` have different meanings.

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -323,20 +333,32 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
...
   delete registration[Constants.ATOM_METADATA_MARKER];
   delete registration.ExpirationTime;
   delete registration.ETag;
 
   if (!registration.Expiry) {
-    delete registration.Expiry;
+      delete registration.Expiry;
+  }
+
+  if (registration.RegistrationId != null) {
```
amarzavery: same as above

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -323,20 +333,32 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
...
+      delete registration.RegistrationId;
   }
 
-  registration.BodyTemplate = '<![CDATA[' + registration.BodyTemplate + ']]>';
+    var newRegistration = {};
+    newRegistration.Tags = registration.Tags;
+    for (var key in registration)
+    {
+        if(key!="Tags")
```
amarzavery: spacing is incorrect and use !==

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -323,20 +333,32 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
...
+        newRegistration[key] = registration[key];
+    }
+
+    if(registration.BodyTemplate!=null)
```
amarzavery: spacing is incorrect. Check for both null and undefined.

lib/services/serviceBus/package.json
amolr: Made the change in other pull request

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -290,14 +290,14 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
       delete registration.Expiry;
   }
 
-  if (registration.RegistrationId != null) {
+  if (registration.RegistrationId !== null || registration.RegistrationId !== undefined) {
```
amarzavery: Actually I played around with the delete operator to find out what happens when you try to delete a property from an object, when the property does not exist in the object. It turns out, it is completely fine  as you can see below.

```bash
node
> obj = {}
{}
> delete obj.foo
true
> obj.foo = null
null
> obj
{ foo: null }
> delete obj.foo
true
> obj.foo = undefined
undefined
> obj
{ foo: undefined }
> delete obj.foo
true
```
However, one needs to make sure that the object is actually defined before executing `delete`
```bash
> delete baz.foo
ReferenceError: baz is not defined
    at repl:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:121:20)
    at REPLServer.emit (events.js:211:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)
```

So, I think we should simply do the following:

```javascript
if (registration) delete registration.registrationId;
```

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -343,19 +343,19 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
...
     newRegistration.Tags = registration.Tags;
     for (var key in registration)
     {
-        if(key!="Tags")
+        if(key !== 'Tags')
         newRegistration[key] = registration[key];
     }
 
-    if(registration.BodyTemplate!=null)
+    if(registration.BodyTemplate !== null || registration.BodyTemplate !== undefined)
```
amarzavery: This should have a **logical AND**. Basically we want to check that if registration object has BodyTemplate property then we want to do something with newRegistration. Based on the current logic, even when registration.BodyTemplate is null or undefined, the if condition will pass.

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -343,19 +343,19 @@ NotificationHubService.prototype.createOrUpdateRegistration = function (registra
       delete registration.Expiry;
   }
 
-  if (registration.RegistrationId != null) {
+  if (registration.RegistrationId !== null || registration.RegistrationId !== undefined) {
```
amarzavery: same as above.

amolr:
@amarzavery addressed the changes mentioned

lib/services/serviceBus/lib/notificationhubservice.js
```diff
@@ -282,16 +282,26 @@ NotificationHubService.prototype.updateRegistration = function (registration, op
...
 
-  var registrationXml = registrationResult.serialize(registrationType, registration);
+  if (registration) {
+      delete registration.RegistrationId;
+    }
+
+    var newRegistration = {};
+    newRegistration.Tags = registration.Tags;
+    for (var key in registration) {
+        if (key !== 'Tags')
```
amarzavery: @amolr - Indentation in javascript is 2 spaces.

lib/services/serviceBus/package.json
amarzavery: @amolr - Can you update `"azure-common": "^0.9.19"` dependency in this PR. I have closed the other PR.

PR #2454, opened 2018-03-12
multiple auth related changes

amarzavery:


amarzavery:
@yugangw-msft - Can you please review this PR?

runtime/ms-rest-azure/lib/credentials/msiTokenCredentials.js
```diff
@@ -36,16 +36,55 @@ class MSITokenCredentials {
...
+        // possibly a Date string '09/14/2017 00:00:00 PM +00:00'
+        if (parsedBody['expires_on'].includes(':') || parsedBody['expires_on'].includes('/')) {
+          parsedBody.expiresOn = new Date(parsedBody['expires_on'], 10);
+        } else {
+          // normal number as a string '1504130527'
+          parsedBody.expiresOn = new Date(parseInt(parsedBody['expires_on'], 10));
+        }
+      }
+      delete parsedBody['expires_on'];
+    }
```
williexu: If different services return different MSI responses, wouldn't it be preferable to address each response in the corresponding MSICredentials?
i.e. define seperate `parseTokenResponse()` for each MSI response.

Seems to me like trying to accommodate the different logic for every msi response (and new ones in the future) will be harder to maintain.

williexu: After our talk, I'm okay with this for now given we only have to deal with two cases.
Depending on the new MSI responses in the future, we would ideally separate the common logic and the service specific logic in the corresponding parent and children classes.

PR #2466, opened 2018-03-16
BatchAI. SDK for API version 2018-03-01

AlexanderYukhanov:
* added support for job level mounting
* added support for environment variables with secret values
* added support for performance counters reporting in Azure Application Insights
* added support for custom images
* added support for pyTorch deep learning framework
* added API for usage and limits reporting
* added API for listing job files in subdirectories
* now user can choose caching type during NFS creation
* get cluster now reports a path segment generated for storing start task output logs
* get job now reports a path segment generated for job's output directories
* renamed EnvironmentSetting to EnvironmentVariable


lib/services/batchaiManagement/package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "ms-rest": "^2.2.5",
```
amarzavery: ms-rest : ^2.3.2
ms-rest-azure: ^2.5.5

AlexanderYukhanov: Thank you, updated

amarzavery:
@AlexanderYukhanov - We just shipped a new version (2.1.32) of the [autorest.nodejs codengenerator](https://www.npmjs.com/package/@microsoft.azure/autorest.nodejs) which has [added some workarounds](https://github.com/Azure/autorest.nodejs/pull/28#issue-175593968) to the way we export the client from the main (client).js file. This change would make it easy for typescript users to get the model interfaces very easily.

Please execute `autorest --latest` and then regenerate with `gulp codegen --project batchai`. This should have no changes in the method signatures, class definition or model definitions. You would see a change in the way the client is exported in the .js and the .d.ts file.

AlexanderYukhanov:
Thanks for the info, regenerated the SDK

PR #2468, opened 2018-03-19
update AKV dataplane

Lusitanian:
regen code against latest keyvault_preview REST spec, bump package version


lib/services/keyVault/lib/keyVaultClient.d.ts
```diff
@@ -2911,9 +2913,9 @@ declare class KeyVaultClient extends AzureServiceClient {
    *
    *                      {http.IncomingMessage} [response] - The HTTP Response stream if an error did not occur.
    */
-  setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, options?: { customHeaders? : { [headerName: string]: string; } }): Promise<models.Contacts>;
-  setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, callback: ServiceCallback<models.Contacts>): void;
-  setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, options: { customHeaders? : { [headerName: string]: string; } }, callback: ServiceCallback<models.Contacts>): void;
+  setCertificateContacts(vaultBaseUrl: string, options?: { contactList? : models.Contact[], customHeaders? : { [headerName: string]: string; } }): Promise<models.Contacts>;
```
amarzavery: Looks like a breaking change. The major version of the package should go up.

amarzavery: looks like required parameter contacts was renamed as contactList and has been made optional.

schaabs: @amarzavery we didn't make any changes to these methods or models.  Could this be due to parameter flattening?  I've seen some similar issues like this in other languages where we were able to fix this with extra autorest parameters.

Lusitanian: @amarzavery see what @schaabs said above - do you think this could have something to with autorest configuration? otherwise not sure why it's generating differently

RikkiGibson: I'll follow up on this-- can you share the AutoRest invocation that produced this change?

RikkiGibson: When I pull in the latest gulpfile.js from master and `gulp codegen --project keyvault` I see this signature:

```js
setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, options?: { customHeaders? : { [headerName: string]: string; } }): Promise<models.Contacts>;
setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, callback: ServiceCallback<models.Contacts>): void;
setCertificateContacts(vaultBaseUrl: string, contacts: models.Contacts, options: { customHeaders? : { [headerName: string]: string; } }, callback: ServiceCallback<models.Contacts>): void;
```

This is the signature we actually want, right? The raw AutoRest invocation for reference:
```sh
autorest \
  --output-folder=/Users/rigibson/src/azure-sdk-for-node/lib/services/keyvault/lib \
  --package-name=azure-keyvault \
  --nodejs \
  --license-header=MICROSOFT_MIT_NO_VERSION  \
  --azure-arm \
  https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/keyvault/data-plane/readme.md \
  --tag=package-7.0-preview
```

schaabs: @RikkiGibson this explains the discrepancy.  We are generating the code from the keyvault_preview branch which does specify flattening threshold in the readme.nodejs.md whereas the one in master does not.  I'll update the readme in this branch and we'll regen the code and update the pull request.  Thanks for the help.

schaabs: @Lusitanian I've merged the updated readme's from master back into the keyvault_preview branch, can you regen the code and update the pullrequest?

lib/services/keyVault/lib/keyVaultClient.d.ts
```diff
@@ -5526,7 +5717,7 @@ declare class KeyVaultClient extends AzureServiceClient {
    *
    * @reject {Error|ServiceError} - The error object.
    */
-  purgeDeletedStorgeAccountWithHttpOperationResponse(vaultBaseUrl: string, storageAccountName: string, options?: { customHeaders? : { [headerName: string]: string; } }): Promise<HttpOperationResponse<void>>;
+  purgeDeletedStorageAccountWithHttpOperationResponse(vaultBaseUrl: string, storageAccountName: string, options?: { customHeaders? : { [headerName: string]: string; } }): Promise<HttpOperationResponse<void>>;
```
RikkiGibson: Is this a breaking change?

schaabs: @RikkiGibson yes this is an intentional breaking change. This API is only in our preview release, so we are fixing the typo before we GA.

RikkiGibson: Sounds good 👍

PR #2485, opened 2018-03-20
Rename filesystem to fileSystem

SoundBot:
filesystem property doesn't exist, but fileSystem works for me


daschult:
Thanks for your contribution!

PR #2506, opened 2018-03-21
Add datamigration management library

daschult:


gulpfile.js
```diff
@@ -329,6 +336,8 @@ gulp.task('sync-mappings-with-repo', (cb) => {
         if (fs.existsSync(rm)) {
           mappings[rp]['resource-manager'] = {
             "packageName": `azure-arm-${rp.toLowerCase()}`,
+            "packageVersion": "1.0.0-preview",
+            "generateMetadata": true,
```
RikkiGibson: Have we decided that there are some projects where we always want to recreate the package.json and readme files?

daschult: This is my vote. I think it would be good if we could always generate the package.json and README.md files. That way we never have to do any manual changes after generating, which can easily cause bugs. I think we should aim to push a single button which will generate all packages and then (eventually) publish them too. We can talk more about the publishing portion, but I think having a generated packageVersion will move us in that direction.

RikkiGibson: I can buy it for package.json where things should be very predictable. It seems like README.md may have some bona-fide usage documentation that goes beyond the generated sample, though.

RikkiGibson: Perhaps that's simply something to manage with git, and any changes made to the README by the automated system should be at least looked at.

RikkiGibson:
What's up with CI?

daschult:
There's a legitimate problem either with the swagger or with our generator. I need to talk to Amar about this.

daschult:
Yup, the CI is failing because the datamigration swagger is wrong. I'll send an e-mail to that team and ask them to fix their swagger.

daschult:
This PR should not be merged until https://github.com/Azure/azure-rest-api-specs/issues/2728 is fixed. Once the issue has been resolved, then this SDK should be generated again and the PR should be updated.

PR #2513, opened 2018-03-22
Add codegen_mappings schema file and small gulpfile tweak

daschult:
Adding a JSON schema for codegen_mappings.json doesn't add any functionality, but it does add editor IntelliSense and descriptions. The descriptions are always a good reminder for what a property is used for.
If we add more properties to each of the package objects within codegen_mappings.json, a schema is a great way to ensure that you didn't forget to add required properties to any packages.

### DIFF
--- a/codegen_mappings.schema.json
+++ b/codegen_mappings.schema.json
@@ -1,0 +1,167 @@
+{
+    "type": "object",
+    "properties": {
+        "$schema": {
+            "type": "string"
+        }
+    },
+    "additionalProperties": {
+        "description": "An Azure service.",
+        "type": "object",
+        "properties": {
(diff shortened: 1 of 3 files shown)


codegen_mappings.schema.json
```diff
@@ -0,0 +1,167 @@
...
+                    "additionalProperties": {
+                        "$ref": "#definitions/resourceManagerPackage"
+                    }
+                }
+            ]
+        },
+        "resourceManagerPackage": {
+            "description": "The management package for this service.",
+            "type": "object",
+            "properties": {
```
RikkiGibson: Is this identical to the dataplane package except the description?

PR #2531, opened 2018-03-22
Regenerate compute management SDK to support Availability Zones

daschult:


RikkiGibson:
Looks like test rerecordings are needed? Also, man, a 26,000 line operations.d.s file, that's hefty.

daschult:
@RikkiGibson I guess you commented earlier, so could you sign off on this one?

PR #2617, opened 2018-03-30
Remove codegen_mappings.json, update gulpfile tasks

daschult:
In this PR, I've moved all of the contents of codegen_mappings.json to the readme.nodejs.md files in the azure-rest-api-specs repository (see https://github.com/Azure/azure-rest-api-specs/pull/2788 for the results of the move). This means that all of the new swagger specification PRs will have the latest information needed to generate Node.js packages. This also means that since the details are now in the azure-rest-api-specs repository in readme.md files, we can just point autorest at those readme files. This means that we don't need codegen_mappings.json any more. This PR also updates our gulpfile to generate our packages by iterating over your local clone of the azure-rest-api-specs repo and looking for readme.nodejs.md files. It will generate a package for each of those files that it finds.


gulpfile.js
```diff
@@ -6,245 +6,131 @@
...
+          nodejsReadmeFilePaths.push(folderEntryPath);
+        }
       }
-    });
-    fs.rmdirSync(path);
+    }
   }
-}
 
-function clearProjectBeforeGenerating(projectDir) {
```
amarzavery: W.r.t cleaning projects before generating there are sdks like azure-batch, azure-keyvault and datalake analytics/filesystem that have some hand written files along with the autogenerated files. We should make sure that Autorest does not accidentally delete them.

daschult: I'll verify this. Hold on.

gulpfile.js
```diff
@@ -436,102 +234,26 @@ gulp.task('sync-deps-rollup', (cb) => {
   fs.writeFileSync('./package.json', JSON.stringify(rollupPackage, null, 2), { 'encoding': 'utf8' });
 });
 
-gulp.task('sync-package-service-mapping', (cb) => {
```
amarzavery: this task is used to update package_service_mapping.json . This file is used by the docs team to update sdk docs on docs.microsoft.com website. We need to make sure that we update this mapping once we publish packages. I think I missed letting you know about this. It is my fault.

gulpfile.js
daschult: Yup, I checked and the handwritten stuff stayed put.

gulpfile.js
daschult: Don't worry about it. It's not possible to tell me everything in advance, and even if you tried, I probably wouldn't remember it. This is the perfect time to learn about this. I'll fix it.

codegen_mappings.schema.json
```diff
@@ -1,197 +0,0 @@
-{
```
RikkiGibson: Goodbye codegen_mappings.schema.json, we hardly knew ye.

gulpfile.js
```diff
@@ -6,245 +6,131 @@
...
-    console.log(execHelpCmd);
-    execHelp = execSync(execHelpCmd, { encoding: 'utf8' });
-    //console.debug(execHelp);
-  } catch (err) {
-    result = false;
-    console.log(`An error occurred while getting the "${version}" of autorest and executing "autorest --help":\n ${util.inspect(err, { depth: null })}.`);
-  }
-  return result;
-}
+function findReadmeNodejsMdFilePaths(azureRestAPISpecsRoot) {
```
RikkiGibson: Can't we use something like node-glob for this?

gulpfile.js
```diff
@@ -6,245 +6,131 @@
...
-        }
-        generateProject(proj[key], specRoot, usingAutoRestVersion);
-      } else {
-        iterateProject(proj[key], specRoot, usingAutoRestVersion);
+      try {
+        console.log('Executing command:');
+        console.log('------------------------------------------------------------');
+        console.log(cmd);
+        console.log('------------------------------------------------------------');
+        const result = execSync(cmd, { encoding: 'utf8' });
```
RikkiGibson: This PR might not be the time, but I'd really like to drill into why we can't do generation in parallel.

daschult: I'm sure we can, but you're totally right that this isn't the PR for it.

gulpfile.js
daschult: Yeah, we probably can. I can look into that.

daschult: Done.

gulpfile.js
daschult: Done.

PR #2619, opened 2018-03-30
Update README.md

dend:
Preparing for integration to docs.microsoft.com

### DIFF
--- a/lib/services/advisorManagement/README.md
+++ b/lib/services/advisorManagement/README.md
@@ -1,3 +1,8 @@
+---
+uid: azure-arm-advisor
+summary: *content
+---
+
 # Microsoft Azure SDK for Node.js - AdvisorManagement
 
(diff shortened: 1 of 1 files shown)


daschult:
This looks fine, but for this to have a lasting change it'll have to be put into the generator. Can you explain what the different parts of your change do?

dend:
@daschult this flags the content as part of package documentation, that we can then render under the appropriate package page on docs.microsoft.com. The YAML front-matter should not be affecting any product functionality.

daschult:
Okay, cool. Is this pretty stable or are you still working on the contents of the YAML block?

dend:
@daschult this should be stable.

daschult:
@dend Okay. I'll merge this and then send a PR with the generator changes so you can make sure that I got them correct.

dend:
@daschult sounds good, thanks!

PR #2659, opened 2018-04-04
Regenerate computer vision and change version to 1.0.1

daschult:


lib/services/computerVision/package.json
```diff
@@ -1,35 +1,21 @@
...
-  ],
-  "main": "./lib/computerVisionAPIClient.js",
-  "types": "./lib/computerVisionAPIClient.d.ts",
-  "license": "MIT",
+  "description": "ComputerVisionAPIClient Library with typescript type definitions for node",
+  "version": "1.0.1",
   "dependencies": {
-    "ms-rest": "^2.2.5",
-    "ms-rest-azure": "^2.5.0"
+    "ms-rest": "^2.3.2"
```
RikkiGibson: No ms-rest-azure is intended right?

daschult: Good question. I'll double-check.

daschult: Yup, that's intentional.

PR #2680, opened 2018-04-04
 Run 'npm install' after generating and before publishing

daschult:


gulpfile.js
```diff
@@ -39,6 +39,10 @@ function getServiceNameFromOutputFolderValue(outputFolderValue) {
   return outputFolderSegments[outputFolderSegments.length - 1];
 }
 
+function npmInstall(packageFolderPath) {
+  execSync(`npm install`, { cwd: packageFolderPath, stdio: ['ignore', 'ignore', 'pipe'] });
```
RikkiGibson: does this make it so we only see errors from the child process?

daschult: Yes.

PR #2706, opened 2018-04-09
Fix proper Data Lake Store naming

dend:


dend:
Ping @amarzavery @daschult

PR #2746, opened 2018-04-16
Add --package parameter to gulp publish task

daschult:


gulpfile.js
```diff
@@ -258,33 +262,36 @@ gulp.task('publish-packages', (cb) => {
...
+            errorPackages++;
           }
           else {
-            console.log(`Publishing package "${packageName}" with version "${localPackageVersion}"...`);
+            let npmPackageVersion;
             try {
-              npmInstall(packageFolderPath);
-              execSync(`npm publish`, { cwd: packageFolderPath });
-              publishedPackages++;
+              const npmViewResult = JSON.parse(execSync(`npm view ${packageName} --json`, { stdio: ['pipe', 'pipe', 'ignore'] }));
```
RikkiGibson: Nice, so this takes care of the "check if an up to date package was published" thing?

daschult: Yeah. We only attempt to publish if we need to.

PR #2805, opened 2018-04-27
Update EventGrid Management SDK to 1.0.0

RikkiGibson:


lib/services/eventgridManagement/lib/models/topic.js
```diff
@@ -43,30 +43,6 @@ class Topic extends models['TrackedResource'] {
...
-            }
-          },
-          type: {
-            required: false,
-            readOnly: true,
-            serializedName: 'type',
-            type: {
-              name: 'String'
-            }
-          },
```
daschult: It looks like when the base type changed from `Resource` to `BaseResource` we lost these properties. If that's the case, then this is a breaking change and needs a major version bump.

lib/services/eventgridManagement/package.json
```diff
@@ -2,9 +2,9 @@
   "name": "azure-arm-eventgrid",
   "author": "Microsoft Corporation",
   "description": "EventGridManagementClient Library with typescript type definitions for node",
-  "version": "1.2.0",
+  "version": "1.1.0",
```
daschult: The version number can't go backwards. Remember to increase the version number in the readme.nodejs.md file. Double-check on my comment above to determine whether this version should be `2.0.0` or `1.3.0`.

PR #2883, opened 2018-05-21
Storage/SAS definition tests + test recordings for Key Vault

Lusitanian:


test/services/keyVault/keyVault-storage-tests.js
```diff
@@ -0,0 +1,228 @@
...
+var acclocation;
+var akvClient;
+var testVault;
+var userOid;
+var testStorageAccount;
+
+var _tokenFromUserCredentials = () => { 
+  return new Promise ( (resolve, reject) => { 
+    if(suite.isPlayback) {
+      resolve( { 'oid': '7541419d-883d-452f-a823-56aa8bf0749f' } );
```
schaabs: >resolve( { 'oid': '7541419d-883d-452f-a823-56aa8bf0749f' } ); [](start = 6, length = 61)

you should avoid hardcoding values here that might change when the tests are re-recorded.  possibly drop to a file when running live and read when playback?

PR #2917, opened 2018-05-29
lib/credentials/msiVmTokenCredentials.js requires request

peterromfeldhk:
request required without explicit listed in package.json.
which can cause issues with other buildtools
```
Error: Cannot find module 'request'

```

### DIFF
--- a/runtime/ms-rest-azure/package.json
+++ b/runtime/ms-rest-azure/package.json
@@ -29,7 +29,8 @@
     "uuid": "^3.2.1",
     "adal-node": "^0.1.27",
     "ms-rest": "^2.3.2",
-    "moment": "^2.20.1"
+    "moment": "^2.20.1",
+    "request": "^2.83.0"
   },
   "devDependencies": {
     "@types/mocha": "^2.2.40",


runtime/ms-rest-azure/package.json
```diff
@@ -29,7 +29,8 @@
     "uuid": "^3.2.1",
     "adal-node": "^0.1.27",
     "ms-rest": "^2.3.2",
-    "moment": "^2.20.1"
+    "moment": "^2.20.1",
+    "request": "^2.83.0"
```
amarzavery: This should not be the case. What version of npm are you using? 
`ms-rest-azure -(depends-on)-> ms-rest -(depends-on)->request`. Newer versions of npm dedupe/flatten the dependency structure. Are you using request somewhere in your app? If so then it may have a different version which would be colliding with the version ms-rest depends on. As a result npm may be installing ms-rest inside ms-rest-azure, or request under ms-rest, rather than a top level folder inside node_modules.

peterromfeldhk:
its a custom buildscript, we worked around with overriding your
package.json during build..

besides that i also dont think you want to depend on a dependency off a
dependency for something you directly require in your code

PR #2945, opened 2018-06-05
add azure stack env

binbinlu:


runtime/ms-rest-azure/lib/azureEnvironment.js
```diff
@@ -124,6 +124,25 @@ var supportedEnvironments = {
     // TODO: add dns suffixes for the US government for datalake store and datalake analytics once they are defined.
     azureDataLakeStoreFileSystemEndpointSuffix: 'N/A',
     azureDataLakeAnalyticsCatalogAndJobEndpointSuffix: 'N/A'
+  },
+  AzureStack: {
+    name: 'AzureStack',
+    portalUrl: 'http://portal.microsoftazure.de/',
```
amarzavery: How can the portal url be that of German cloud? This seems incorrect to me.

runtime/ms-rest-azure/lib/azureEnvironment.js
```diff
@@ -124,6 +124,25 @@ var supportedEnvironments = {
...
+    publishingProfileUrl: 'N/A',
+    managementEndpointUrl: 'https://management.northwest1.azurestack.selfhost.corp.microsoft.com',
+    resourceManagerEndpointUrl: 'https://management.northwest1.azurestack.selfhost.corp.microsoft.com',
+    sqlManagementEndpointUrl: 'N/A',
+    sqlServerHostnameSuffix: 'N/A',
+    galleryEndpointUrl: 'N/A',
+    activeDirectoryEndpointUrl: 'https://login.windows.net/',
+    activeDirectoryResourceId: 'https://management.msazurestack.onmicrosoft.com/8bf87789-7935-4e97-81f8-55959c31f4dd',
+    activeDirectoryGraphResourceId: 'https://graph.windows.net/',
```
amarzavery: Are you sure that the `activeDirectoryEndpointUrl` and `activeDirectoryGraphResourceId ` are correct? Would they end in `*.windows.net/`?

amarzavery:
@yugangw-msft - Can you please review this change or point me to a place where you have endpoints specified for Azure Stack in the azure-cli or python ms-rest. I would like to make sure that the values are consistent.

amarzavery:
I don't see any mention of AzureStack in azure-cli over [here](https://github.com/Azure/azure-cli/blob/master/src/azure-cli-core/azure/cli/core/cloud.py#L148-L220). If this is not supported in Azure cli then why are we adding this in the nodesdk runtime? Am I missing something?

runtime/ms-rest-azure/lib/azureEnvironment.js
binbinlu: oops... apparently it is not used in test

runtime/ms-rest-azure/lib/azureEnvironment.js
binbinlu: yes

runtime/ms-rest-azure/lib/azureEnvironment.js
```diff
@@ -124,6 +124,25 @@ var supportedEnvironments = {
     // TODO: add dns suffixes for the US government for datalake store and datalake analytics once they are defined.
     azureDataLakeStoreFileSystemEndpointSuffix: 'N/A',
     azureDataLakeAnalyticsCatalogAndJobEndpointSuffix: 'N/A'
+  },
+  AzureStack: {
+    name: 'AzureStack',
+    portalUrl: 'http://onpremportalrc.northwest3.cloudapp.azurestack.selfhost.corp.microsoft.com/',
```
bganapa: binbin why are hardcoding the URLs here? for azure stack, the URLs  will be different for each environment. You should use the metadata endpoint to retrive these values rather than hard coding it

bganapa:
@yugangw-msft , @amarzavery , @binbinlu  Python SDK/Ruby SD/Go autorest  provides a method to init the azurestack cloud reading the values from the metadata endpoint. We should have similar method here rather than hard coding the values. I will talk to binbin offline and understand the scenario here and get back..

e.g. python method
https://github.com/Azure/msrestazure-for-python/blob/master/msrestazure/azure_cloud.py#L229

PR #2984, opened 2018-06-12
Validate authority

### LINKED ISSUE #2952: Typescript: AzureEnvironment.Azure is not assignable to type AzureEnvironment
The type definition for the specific Azure instances are missing the non-optional `validateAuthority` boolean:

https://github.com/Azure/azure-sdk-for-node/blob/8e520cb938edc7bad272163ae8c6bf50f35a5934/runtime/ms-rest-azure/index.d.ts#L257

https://github.com/Azure/azure-sdk-for-node/blob/8e520cb938edc7bad272163ae8c6bf50f35a5934/runtime/ms-rest-azure/index.d.ts#L283-L301

As a result, this code fails to compile:
```typescript
let azureAuth = await msrest.loginWithServicePrincipalSecretWithAuthResponse(
    clientId,
    clientSecret,
    domain,
    {
        environment: msrest.AzureEnvironment.Azure,
    },
);}
```

the error is:
`Property 'validateAuthority' is missing in type '{ name: "Azure"; portalUrl: "https://portal.azure.com"; publishingProfileUrl: "https://go.microso...'.`

RikkiGibson:
Fixes #2952

`validateAuthority` was missing from the type definitions. Because we want to make it easy for users to see what values are used in a given AzureEnvironment, we're using the string and boolean literal types and building something that needs to be assignable to AzureEnvironment.

### DIFF
--- a/runtime/ms-rest-azure/Changelog.md
+++ b/runtime/ms-rest-azure/Changelog.md
@@ -1,3 +1,9 @@
+### 2.5.7 (06/12/2018)
+- Added `validateAuthority` to AzureEnvironment type definitions
+
+### 2.5.6 (06/12/2018)
+- Updated vulnerable dependencies
+
 ### 2.5.5 (03/12/2018)
 - Added an `interface TokenResponse` that extends `adal.TokenResponse`. This interface would be the return type of `getToken()` method on the credential classes.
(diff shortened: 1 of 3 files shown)


RikkiGibson:
Thanks for bringing this issue to our attention @mmdriley. It should be fixed now in `ms-rest-azure@2.5.7`.

PR #2994, opened 2018-06-13
Add additionalInfo to CloudError

j5lim:
'additionalInfo', a new property added to ARM error contract, contains additional error information. For example, it will contain detailed policy violation error information when the request to ARM failed due to violation of Azure policies.

The changes are to allow deserializing additionalInfo in ARM error based on the error type.


j5lim:
@amarzavery Could you please review the PR?

runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -108,6 +114,20 @@ class CloudError extends Error {
...
+            type: {
+              name: 'Sequence',
+              element: {
+                required: false,
+                serializedName: 'AdditionalInfoElementType',
+                type: {
+                  name: 'Object'
+                }
+              }
+            }
```
daschult: How do we expect end-users to use this information? Do we really expect them to just output it to log, such as `console.log(JSON.stringify(cloudError.additionalInfo))`? If you expect them to perform any type of logic/decisions on this data, then I'd recommend coming up with a better type definition than just Object. If you just expect `additionalInfo` to be output to a log, then this is probably fine.

runtime/ms-rest-azure/lib/cloudError.js
```diff
@@ -108,6 +114,20 @@ class CloudError extends Error {
...
+          additionalInfo: {
+            required: false,
+            serializedName: 'additionalInfo',
+            type: {
+              name: 'Sequence',
+              element: {
+                required: false,
+                serializedName: 'AdditionalInfoElementType',
+                type: {
+                  name: 'Object'
```
amarzavery: I believe the structure of objects inside the additionalnfo array should be like this:
```js
type: {
  name: 'Composite'
  className: 'AdditionalErrorInfo'
  modelProperties: {
    type: {
      required: true,
      serializedName: 'type',
      type: {
          name: 'String'
      }
    },
    info: {
      required: true,
      serializedName: 'info',
       type: {
         name: 'Object'
       }
    }
  }
}
```
[.net sdk runtime](https://github.com/Azure/azure-sdk-for-net/blob/psSdkJson6/src/SdkCommon/ClientRuntime.Azure/ClientRuntime.Azure/CommonModels/AdditionalErrorInfo.cs) and [python sdk runtime](https://github.com/Azure/msrestazure-for-python/blob/master/msrestazure/azure_exceptions.py#L233) have similar type definitions.

runtime/ms-rest-azure/lib/cloudError.js
j5lim: Defined 'type' and 'info' property as amarzavery suggested. I expect the users to usually log these additional error information. Also 'type' property will tell the error type which I think can be used for any further decision.

runtime/ms-rest-azure/lib/cloudError.js
j5lim: Thanks. Updated the definition.

PR #2997, opened 2018-06-14
update package.json and publish graph 3.0.0

amarzavery:


amarzavery:
@RikkiGibson and @daschult - Looks like we generated the graph library two months ago but for some reason did not publish it. I am updating the package dependencies and publishing a new version as a partner team needs it.

PR #3048, opened 2018-06-21
x-nullable

### LINKED ISSUE #3015: Support XNullable in Node.js SDK
- We need to add support for `x-nullable: true` in the generator and in the serializer as well. We need to add a check over [here](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/lib/serialization.js#L231) to make sure that we don't throw error if the property is nullable.
- Add a test if there does not exist one where we want to set a property of a composite model to null and that property with null value should be present in the request body on the wire

RikkiGibson:
Fixes #3015

### DIFF
--- a/runtime/ms-rest/Changelog.md
+++ b/runtime/ms-rest/Changelog.md
@@ -1,3 +1,6 @@
+### 2.3.5 (6/21/2018)
+- Add support for x-nullable Swagger extension
+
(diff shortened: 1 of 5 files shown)


RikkiGibson:
Still trying to figure out if this works will all the autorest.nodejs tests.

PR #3061, opened 2018-06-25
Regenerate sql

RikkiGibson:
Testing the new nullability feature in the generator. Perhaps we want to bump version and ship this?


lib/services/sqlManagement2/package.json
```diff
@@ -7,7 +7,10 @@
     "ms-rest": "^2.3.3",
```
daschult: This requires at least a major version bump.

RikkiGibson: We are moving from '2015-05-01-preview' to '2017-03-01-preview', so yeah let's do that.

RikkiGibson:
Published `azure-arm-sql@4.0.0`

PR #3119, opened 2018-07-05
Generating mediaservices for 2018-06-01-preview.

amarzavery:


lib/services/mediaServicesManagement/lib/models/index.d.ts
```diff
@@ -1936,20 +1939,19 @@ export interface NoEncryption {
...
+ * separate content key
+ * @member {string} [customKeyAcquisitionUrlTemplate] KeyAcquistionUrlTemplate
+ * is used to point to user specified service to delivery content keys
  */
 export interface EnvelopeEncryption {
   enabledProtocols?: EnabledProtocols;
   clearTracks?: TrackSelection[];
   contentKeys?: StreamingPolicyContentKeys;
-  customLicenseAcquisitionUrlTemplate?: string;
+  customKeyAcquisitionUrlTemplate?: string;
```
amarzavery: This is a breaking change. Hence bumping the sdk version from 2.0.0 to 3.0.0

PR #3160, opened 2018-07-11
Regenerate custom vision prediction package

daschult:


daschult:
@RikkiGibson This is a package that I missed. CustomVision has two packages (training and prediction). This is prediction.

daschult:
Published to https://www.npmjs.com/package/azure-cognitiveservices-customvision-prediction/v/1.0.0-preview.

PR #3208, opened 2018-07-20
Generate AutoSuggest data-plane library

daschult:


daschult:
@RikkiGibson Yeah, it's wrong too. Most of those parameters are optional, which means they should be in an options object. We need to update the readme generator so that it only puts out required parameters.

PR #3233, opened 2018-07-25
Merge websiteManagement2 and webSiteManagement2 folders

daschult:
We got into a weird situation where we had both a websiteManagement2 folder and a webSiteManagement2 folder, which was causing problems on Windows machines. This PR should solve that problem.


daschult:
Since we talked about it yesterday ("Website hasn't been two words since the 90's." seems to be the sentence that really sticks out to me) I agreed that website shouldn't be capitalized. This PR is just about deleting the capitalized version and going to all lowercaps. It took much more effor that it should've, but at least it is done now.

daschult:
Especially once https://github.com/Azure/azure-rest-api-specs/pull/3499 gets merged.

PR #3278, opened 2018-08-01
Fixed issue with environment credentials

joecoolish:
We ran into an issue where our Azure US GOV subscription would not be found.  We weren't passing in a new base uri because we assumed that the resource management uri would be taken from the credentials.  This was not the case and it caused some major debugging to occur.

It turns out, that the there was a hard coded "magic string" that caused these issues, and this pull request fixes that problem.

### DIFF
--- a/lib/services/computeManagement2/lib/computeManagementClient.js
+++ b/lib/services/computeManagement2/lib/computeManagementClient.js
@@ -55,7 +55,7 @@ class ComputeManagementClient extends ServiceClient {
     this.generateClientRequestId = true;
     this.baseUri = baseUri;
     if (!this.baseUri) {
-      this.baseUri = 'https://management.azure.com';
+      this.baseUri = credentials.environment.resourceManagerEndpointUrl || 'https://management.azure.com';  // Should pull the base URI from the environment of the credentials by default
     }
     this.credentials = credentials;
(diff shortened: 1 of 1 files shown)


RikkiGibson:
Related to #3280

Thanks for your contribution @joecoolish! There are just some minor changes to the implementation I'd like to make here:

- computeManagementClient.js is generated code. Let's instead make this change in [AzureServiceClient](https://github.com/joecoolish/azure-sdk-for-node/blob/efe330ca7123eaade394a22551b792958d01e27d/runtime/ms-rest-azure/lib/azureServiceClient.js#L343) so that every management SDK will benefit from this.
- Not all credentials objects have an environment property so let's do something like:
```js
this.baseURL = credentials.environment
  ? credentials.environment.resourceManagerEndpointUrl
  : "https://management.azure.com"
```

@amarzavery can we assume in general that the resourceManagerEndpointUrl from the environment is the one we want to use? I guess we can still have people explicitly pass the baseURL with another property off the environment if they need a different endpoint?

RikkiGibson:
Spoke with Amar about it for a bit, and we think it's fine to always use `.resourceManagerEndpointUrl` in all cases in the AzureServiceClient.

amarzavery:
closing this PR as the work needs to happen in AzureServiceClient.

PR #3294, opened 2018-08-06
Add App Insights data plane query SDK

alexeldeib:
After some additional offline discussion, I'm closing https://github.com/Azure/azure-sdk-for-node/pull/3021 and opening this PR to add support for App Insights data plane querying using the typescript generator. There wasn't an obvious pattern here, so I mirrored what I saw from https://github.com/amarzavery/azure-batch-ts as much as I could. 

Couple questions came up. I tried out the library and ran into https://github.com/Azure/ms-rest-js/issues/156. I browsed ms-rest-nodeauth and ms-rest-js and decided to try doing the auth with loginWithServicePrincipal, but looks like the types from the auth library and the service client aren't compatible? I'm curious what the intended story is around auth. 

image

I'm also trying to use a custom resource and from the code it looks that that will not be possible using ms-rest-nodeauth with service principal; It currently looks like my available option is using adal-node to acquire tokens and create TokenCredentials? I figure this and the above will probably go hand in hand. 

I'll also clarify examples, readme, etc. 

Thanks!

### DIFF
--- a/lib/services/applicationinsightsQuery/package.json
+++ b/lib/services/applicationinsightsQuery/package.json
@@ -4,8 +4,7 @@
   "description": "ApplicationInsightsDataClient Library with typescript type definitions for node.js and browser.",
   "version": "",
   "dependencies": {
-    "ms-rest-azure-js": "~0.12.123",
-    "ms-rest-nodeauth": "^0.3.0"
+    "ms-rest-azure-js": "~0.12.123"
   },
   "keywords": [
     "node",


RikkiGibson:
ms-rest-nodeauth is the intended way to do ADAL authentication with the new TS SDKs.

It sounds like weird type checking issues are happening around the identity of the types referenced in each definition of ServiceClientCredentials in the conflicting versions of ms-rest-js. I've published `ms-rest-nodeauth@0.3.0` to depend on the newest ms-rest-js preview-- please give it a try.

alexeldeib:
@RikkiGibson The types are still slightly mismatched. Fiddled around a bit, I think the mismatch is from ms-rest-azure-js. Here are the versions of ms-rest-js I see the packages pulling in:

ms-rest-azure-js@0.8.82 => ms-rest-js@0.12.293
ms-rest-nodeuath@0.2.0 => ms-rest-js@0.14.335 
ms-rest-nodeauth@0.3.0 => ms-rest-js@0.17.376

Could you bump ms-rest-azure-js to the same version you did for ms-rest-nodeauth? I also tried using the older ms-rest-nodeauth versions but couldn't find a proper match.

RikkiGibson:
Does it still fail with everything installed `@preview`? Sorry that it's been a little bumpy to get started with--we're expecting to stabilize quite soon.

alexeldeib:
Regenerated with `--use=@microsoft.azure/autorest.typescript@preview` and got the same version as ms-rest-nodeauth@0.3.0 😄 Seems to work okay now. 

I can make calls, but I'm still hitting 403s because of the token resource issue. Is there a way I can add and use a custom token resource?

RikkiGibson:
It sounds like some recent improvements were made in ms-rest-azure and not ported over to ms-rest-nodeauth. We'll do some work to move those improvements over.

RikkiGibson:
@alexeldeib It looks like it will probably take another week for ms-rest-nodeauth to stabilize.

alexeldeib:
@daschult Is this defunct/should I close it out? I saw you generated and merged the [query code](https://github.com/Azure/azure-sdk-for-node/commits/master/lib/services) and [published the package](https://www.npmjs.com/package/azure-applicationinsights-query)?

If so, thank you! 😄

alexeldeib:
@daschult @kpajdzik Could you also clarify the difference between these two folders?
- https://github.com/Azure/azure-sdk-for-node/tree/master/lib/services/applicationinsights-query
- https://github.com/Azure/azure-sdk-for-js/tree/master/packages/%40azure/arm-applicationinsights-query

Former seems to be the source of the npm package. Just wondering what the @azure folder is intended for and why it follows the `arm-` naming policy (which AFAIK, is reserved for management plane).

kpajdzik:
@alexeldeib @Azure folder reflects new name convention for all the Azure packages published on NPM (`@azure/<packagename>`). The `arm-` prefix is a mistake in AppInsights data plane case and I'll fix it before publishing.

kpajdzik:
PR: https://github.com/Azure/azure-rest-api-specs/pull/4244

PR #3454, opened 2018-08-29
Add 'gulp install' command

ghost:
Sometimes when I'm working with SDK generation I have to go into a service's folder, run `npm install` and the go back to the main folder of the repository. This gulp task simplifies that process.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -44,6 +44,10 @@ function npmInstall(packageFolderPath) {
 }
 
 gulp.task('default', function () {
+  console.log('gulp install --package <package name>');
+  console.log('  --package');
+  console.log('    NPM package to run "npm install" on.');
+  console.log();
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -57,11 +61,41 @@ gulp.task('default', function () {
...
+
+    let foundPackage = false;
+
+    for (let i = 0; i < nodejsReadmeFilePaths.length; ++i) {
+      const nodejsReadmeFilePath = nodejsReadmeFilePaths[i];
+
+      const nodejsReadmeFileContents = fs.readFileSync(nodejsReadmeFilePath, 'utf8');
+      const packageName = getPackageNameFromReadmeNodejsMdFileContents(nodejsReadmeFileContents);
+
+      if (package === packageName || packageName.endsWith(`-${package}`)) {
```
kpajdzik: What's the case when second condition is true?

daschult: Our published packages look like `azure-arm-storage`, so the first condition is if I entered `gulp install --package azure-arm-storage` and the second condition is if I entered `gulp install --package storage`

PR #3488, opened 2018-09-04
Regenerate azure-arm-analysisservices and update version to 2.2.2-pre…

daschult:
…view


RikkiGibson:
Did it move from stable to preview on purpose?

daschult:
It shouldn't have. I was trying to do too many things at the same time and must've confused this package.json with another one. Thanks for catching that.

PR #3489, opened 2018-09-04
Regenerate azure-arm-hanaonazure and update version to 1.0.1-preview

daschult:


lib/services/hanaonazureManagement/package.json
```diff
@@ -2,9 +2,9 @@
   "name": "azure-arm-hanaonazure",
   "author": "Microsoft Corporation",
   "description": "HanaManagementClient Library with typescript type definitions for node",
-  "version": "1.0.1",
+  "version": "1.0.1-preview",
```
RikkiGibson: Intended?

daschult: Yes, this one is intendend. :)

PR #3494, opened 2018-09-04
Generate azure-arm-datamigration

daschult:


daschult:
Can't generate this package until datamigration fixes their specification.

PR #3514, opened 2018-09-06
Update publish script to use npm-auth env variable and set access to …

daschult:
…public


gulpfile.js
```diff
@@ -278,6 +278,8 @@ gulp.task('publish', (cb) => {
   let publishedPackages = 0;
   let publishedPackagesSkipped = 0;
 
+  const npmAuth = process.env["npm-auth"];
```
RikkiGibson: where is this set/where does this live?

PR #3516, opened 2018-09-06
Fix npm authentication during CI publish

daschult:


gulpfile.js
```diff
@@ -278,23 +229,28 @@ gulp.task('publish', (cb) => {
...
     if (!fs.existsSync(packageFolderPath)) {
-      console.log(`ERROR: Package folder ${packageFolderPath} has not been generated.`);
+      console.log(`WARNING: Package folder ${packageFolderPath} has not been generated.`);
       errorPackages++;
     }
     else {
       const packageJsonFilePath = `${packageFolderPath}/package.json`;
       if (!fs.existsSync(packageJsonFilePath)) {
-        console.log(`ERROR: Package folder ${packageFolderPath} is missing its package.json file.`);
+        console.log(`WARNING: Package folder ${packageFolderPath} is missing its package.json file.`);
```
amarzavery: not a hard requirement but you can also use console.warn|error|info(). More info over [here](https://nodejs.org/dist/latest-v8.x/docs/api/console.html).

Still keeping the WARNING:  prefix makes sense.

PR #3755, opened 2018-10-03
Define msiEndpoint and msiSecret as optional

irnc:
These parameters are optional because their values are set from environment variables provided by App Service.


kpajdzik:
FYI This is already the case in ms-rest-nodeauth library ([source](https://github.com/Azure/ms-rest-nodeauth/blob/4745cab4ce3bf6ac165ec9488b8bed3c60dad996/lib/login.ts#L826))

irnc:
> This is already the case in ms-rest-nodeauth library

@kpajdzik I am a little bit confused, which package should be preferable used for auth?

PR #4128, opened 2018-11-19
Regenerate azure-arm-resource and bump the SwaggerToSDK generator ver…

daschult:
…sion
Related to https://github.com/Azure/autorest.nodejs/issues/69.


lib/services/resourceManagement/lib/resource/models/genericResource.js
```diff
@@ -131,7 +131,7 @@ class GenericResource extends models['Resource'] {
             required: false,
             serializedName: 'kind',
             constraints: {
-              Pattern: '^[-\w\._,\(\)]+$'
+              Pattern: /^[-\w\._,\(\)]+$/
```
vishwam: LGTM, `new RegExp(/^[-\w\._,\(\)]+$/).test('Microsoft.IoTHub')` returns true, but I don't have the context for the rest of the files.

PR #4129, opened 2018-11-19
Rename dataLake.Store to datalake.Store

daschult:
Looks like both dataLake.Store and datalake.Store existed. On my windows machine I didn't notice, but git noticed. This removes dataLake.Store in favor of the casing used in the azure-rest-api-specs readme.nodejs.md file.


kpajdzik:
Why all those files were removed and no new files were added?

daschult:
@kpajdzik I explained it in this PR's comment. Both dataLake.Store and datalake.Store existed in this repository. This PR is just to remove dataLake.Store and to fix the tests.

daschult:
@amarzavery This is the PR that enables case-insensitive property name lookup during deserialization.

runtime/ms-rest/lib/serialization.js
```diff
@@ -484,6 +484,46 @@ function deserializeDictionaryType(mapper, responseBody, objectName) {
...
 
+/**
+ * Get the value of the property with the provided path. This function will first do a
+ * case-sensitive check for the property, and if that returns undefined then it will do a
+ * case-insensitive check.
+ * @param {object} value The value to get the property value from.
+ * @param {string[]} propertyPath The path to the property.
+ * @returns {any} The value of the property or undefined if it wasn't found.
+ */
+function getPropertyValue(value, propertyPath) {
```
amarzavery: I would recommend writing some unit tests for this scenario to make sure that everything is covered.

Property present with null value
Property present with different casing and the value is null
Property present with non null value
Property present with different casing with a non-null value

Test the above scenarios with flattening involved.

daschult: I added these unit tests in this PR: https://github.com/Azure/azure-sdk-for-node/pull/4147

daschult:
@amarzavery Please take another look.

daschult:
Since I responded to all of @amarzavery feedback, I'm going to merge this PR.

PR #4226, opened 2018-11-30
Azure Pipeline for KeyVault

mikeharder:
Pipeline: https://dev.azure.com/azure-sdk/playground/_build?definitionId=31&_a=summary

### DIFF
--- a/.azure-pipelines/client.test.yml
+++ b/.azure-pipelines/client.test.yml
@@ -15,6 +15,8 @@ jobs:
 
     strategy:
       matrix:
+        Node6:
+          node.version: '6.x'
         Node8:
           node.version: '8.x'
         Node10:
--- a/.azure-pipelines/client.test.yml
+++ b/.azure-pipelines/client.test.yml
@@ -11,7 +11,7 @@ jobs:
       vmImage: ${{ parameters.vmImage }}
 
     variables:
-      path: 'test/services/keyVault'
+      build.path: 'test/services/keyVault'
 
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


.azure-pipelines/client.yml
```diff
@@ -0,0 +1,83 @@
...
+        displayName: 'Install Node.js $(node.version)'
+
+      - task: Npm@1
+        inputs:
+          command: 'custom'
+          customCommand: 'i --package-lock-only'
+          workingDir: '$(build.path)'
+        displayName: 'npm i --package-lock-only'
+
+      - powershell: "npm audit --json | ConvertFrom-Json | Select-Object  -expandProperty 'metadata' | select-object -expandproperty 'vulnerabilities' | %{ if ($_.low + $_.moderate + $_.high + $_.critical -ne 0) { write-error 'FAILED! Your package has vulnerabilities. Check npm audit' }}"
```
kpajdzik: Does Ubuntu image on DevOps come with PowerShell preinstalled? 

Also, nice check. We should port this to every JS project instead of looking for vulnerabilities manually.

.azure-pipelines/client.test.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    pool:
+      vmImage: ${{ parameters.vmImage }}
+
+    variables:
+      build.path: 'test/services/keyVault'
+
+    strategy:
+      matrix:
+        Node8:
```
daschult: Supposedly we still support Node6 in the management plane. You may consider adding that to the matrix.

.azure-pipelines/client.yml
mikeharder: Both Ubuntu and Mac images include PowerShell 6.1 by default:

https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md
https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-Readme.md

We're planning to use PowerShell as our default xplat language for most scripts.

.azure-pipelines/client.test.yml
mikeharder: I'm not sure if Node6 is in the support matrix we're planning for the future, but KeyVault tests pass in Node6 so it's good to keep running them.  Added Node6 to this PR.

PR #4237, opened 2018-12-01
Parameterize source and test paths in pipeline

mikeharder:


.azure-pipelines/client.test.yml
```diff
@@ -38,7 +35,7 @@ jobs:
       - task: Npm@1
         inputs:
           command: 'custom'
-          customCommand: 'run mocha -- $(build.path) --reporter mocha-junit-reporter'
+          customCommand: 'run mocha -- $(test.path) --reporter mocha-junit-reporter'
```
daschult: Where is `test.path` defined? Is this an azure-devops thing?

.azure-pipelines/client.yml
```diff
@@ -20,19 +19,19 @@ jobs:
       - task: Npm@1
         inputs:
           command: 'install'
-          workingDir: '$(build.path)'
+          workingDir: '$(source.path)'
```
daschult: Where is `source.path` defined?

.azure-pipelines/client.test.yml
mikeharder: Correct, `test.path` and `source.path` are variables supplied by the DevOps pipeline.

PR #4400, opened 2018-12-06
Regenerate azure-cognitiveservices-websearch

daschult:


lib/services/cognitiveServicesWebSearch/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "azure-cognitiveservices-websearch",
   "author": "Microsoft Corporation",
-  "description": "WebSearchAPIClient Library with typescript type definitions for node",
-  "version": "1.1.0",
+  "description": "WebSearchClient Library with typescript type definitions for node",
+  "version": "2.0.0",
```
kpajdzik: Which changes are breaking?

daschult: The service client got renamed which is a pretty serious breaking change.

PR #4417, opened 2018-12-10
Made msiEndpoint and msiSecret optional

miroslavvojtus:
The options msiEndpoint and msiSecret should be optional as their are loaded from environment by default and does not have to be present to provide another options (e.g. resource). See [line 35-36](https://github.com/Azure/azure-sdk-for-node/blob/27d21ee19fde8bbec45d0b53bcaaf56e8d1beeb3/runtime/ms-rest-azure/lib/credentials/msiAppServiceTokenCredentials.js#L35).

Currently I have to use it as `await msRestAzure.loginWithAppServiceMSI({ resource: "https://vault.azure.net" } as msRestAzure.MSIAppServiceTokenCredentials);`


adxsdk6:

PR #4431, opened 2018-12-11
chore: Add .gitattributes for line endings

nschonni:
Noticed a bunch of the autopr stuff had line ending issues that caused large diffs. This will make sure they get committed as LF and reduce this diff.
This may cause conflicts with open PRs, so I don't know how you want to handle that. Alternately, this could be done without the normalization and just let it get fixed as things are changed by just adding the .gitattributes


nschonni:
running `git rm -rf --cached .` and `git add .` on windows also appears to have removed the execute bit on a bunch of the files. They don't appear like they should be executable though.

adxsdk6:

amarzavery:
Closing this PR since the repo will be deprecated soon.

PR #4543, opened 2019-01-14
Restapi auto redis/resource manager

kpajdzik:


adxsdk6:

lib/services/rediscachemanagement/package.json
```diff
@@ -1,37 +1,25 @@
 {
   "name": "azure-arm-rediscache",
   "author": "Microsoft Corporation",
-  "contributors": [
-    "Thaniyulla Parambath, Alfan <alfantp@microsoft.com>",
-    "Dacosta, Carlyle <cadaco@microsoft.com>",
-    "Zavery, Amar <amzavery@microsoft.com>"
-  ],
+  "description": "RedisManagementClient Library with typescript type definitions for node",
   "version": "2.1.0-preview",
```
daschult: Need to bump the version (I'm guessing 3.0.0).

lib/services/rediscachemanagement/README.md
```diff
@@ -1,113 +0,0 @@
-# Microsoft Azure SDK for Node.js - Redis
```
daschult: You probably don't want to delete the README.md file.

kpajdzik: I removed the directory and regenerated. Is README written manually?

lib/services/rediscachemanagement/package.json
kpajdzik: Yes, I'm waiting on publishing of autorest.nodejs so I can use the official package to regenerate.

lib/services/rediscachemanagement/README.md
daschult: In this package it seems that it is.

PR #4548, opened 2019-01-15
Update createOrUpdateInstallation documentation

Deagler:
Existing documentation describes the installation parameter as a string. It is actually an installation object.

The correct parameter can be in seen in the [Notification Hub REST API Documentation ](https://docs.microsoft.com/en-us/previous-versions/azure/reference/mt621153(v%3dazure.100)#request-body)

You can see evidence that it's an object [here ](https://github.com/Azure/azure-sdk-for-node/blob/1f7ddb946a5537c43a576d3842867d12b78d80cf/lib/services/serviceBus/lib/notificationhubservice.js#L512) or in the screenshot below:
image

I have tested this in my own application and have confirmed that the parameter described in the REST API Documentation is the correct one.

I have also created a [pull request](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32166) to update the type used in the @types/azure-sb package

### DIFF
--- a/lib/services/serviceBus/lib/notificationhubservice.js
+++ b/lib/services/serviceBus/lib/notificationhubservice.js
@@ -456,13 +456,48 @@ NotificationHubService.prototype._sendNotification = function (payload, optionsO
 };
 
 /**
+ *  @typedef TemplateDescription
+ *  @type {object}
+ *  @property {string} body                         Template for the body of notification payload.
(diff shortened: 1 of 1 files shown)


adxsdk6:

Deagler:
Pinging @amarzavery since you've replied to azure-sb PR's in the past.

PR #4583, opened 2019-01-23
Use matrix for vmImage parallelism

### LINKED ISSUE #4563: Use matrix instead of templates for vmImage parallelism
Example: https://github.com/Azure/amqp-common-js/blob/master/.azure-pipelines.yml

mikeharder:
- Fixes #4563

CC: @Azure/azure-sdk-eng


adxsdk6:

PR #4621, opened 2019-01-31
Add script to print package deprecation message

kpajdzik:
Looks like this:

image


adxsdk6:

PR #4639, opened 2019-02-05
Remove duplicated exports in ms-rest-azure

kpajdzik:


runtime/ms-rest-azure/lib/subscriptionManagement/models/index.js
```diff
@@ -18,11 +18,8 @@ exports.BaseResource = require('../../baseResource');
 exports.CloudError = require('../../cloudError');
 exports.Location = require('./location');
 exports.LocationListResult = require('./locationListResult');
-exports.SubscriptionPolicies = require('./subscriptionPolicies');
```
AlexGhiondea: This one wasn't duplicated, right?

adxsdk6:

runtime/ms-rest-azure/lib/subscriptionManagement/models/index.js
kpajdzik: I sorted them so it's easier to spot the duplicates in the future.

PR #4651, opened 2019-02-07
Remove postinstall script from azure-arm-kusto

kpajdzik:


daschult:
Doesn't it need to be removed from all of the packages, not just ms-rest?

kpajdzik:
> Doesn't it need to be removed from all of the packages, not just ms-rest?

That's removal from azure-arm-kusto, the only one generated with postinstall script so far.

adxsdk6:

PR #4661, opened 2019-02-08
Add method to get queue's dead letter queue

### LINKED ISSUE #4660: [Service Bus] There is no obvious way to retrieve messages from the dead letter queue
Currently, there's no available method on the [`ServiceBusService` object](https://github.com/Azure/azure-sdk-for-node/blob/master/lib/services/serviceBus/lib/servicebusservice.js), not any other mechanism I can find, to read messages from a dead letter queue.

Doing some quick research on the .NET client, it seems that all `QueueClient.FormatDeadLetterPath()` does is simply append `/$DeadLetterQueue`, so I can probably send a quick PR out for this. However, this isn't documented in the [Service Bus dead-letter queue docs](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dead-letter-queues).

MLoughry:
Fixes #4660

### DIFF
--- a/lib/services/serviceBus/lib/servicebusservice.js
+++ b/lib/services/serviceBus/lib/servicebusservice.js
@@ -658,6 +658,16 @@ ServiceBusService.prototype.deleteQueue = function (queuePath, callback) {
 };
 
 /**
+* Formats a queue path to point towards its dead letter queue.
+*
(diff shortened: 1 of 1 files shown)


adxsdk6:

PR #5057, opened 2019-04-05
Remove hard coded string “servicebus.windows.net” and use host name

ShivangiReja:
This pr removes hard coded string “servicebus.windows.net” and add host name.

Some methods(delete, unlock etc) fails for china users because the string “servicebus.windows.net” is hardcoded and used in substring operations, so it forms bad url for china users.

### DIFF
--- a/lib/services/serviceBus/lib/servicebusservicebase.js
+++ b/lib/services/serviceBus/lib/servicebusservicebase.js
@@ -80,7 +80,7 @@ function ServiceBusServiceBase(configOrNamespaceOrConnectionStringOrSettings, ac
       }
     }
 
-    var endpoint = url.format({ protocol: 'https:', port: 443, hostname: configOrNamespaceOrConnectionStringOrSettings + '.' + ServiceClientConstants.CLOUD_ACCESS_CONTROL_HOST });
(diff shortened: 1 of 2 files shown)


adxsdk6:

lib/services/serviceBus/lib/servicebusservicebase.js
```diff
@@ -80,8 +80,8 @@ function ServiceBusServiceBase(configOrNamespaceOrConnectionStringOrSettings, ac
       }
     }
 
-    var endpoint = url.format({ protocol: 'https:', port: 443, hostname: configOrNamespaceOrConnectionStringOrSettings + '.' + ServiceClientConstants.CLOUD_SERVICEBUS_HOST });
-    var stsendpoint = url.format({ protocol: 'https:', port: 443, hostname: acsNamespace + '.' + ServiceClientConstants.CLOUD_ACCESS_CONTROL_HOST });
+    var endpoint = url.format({ protocol: 'https:', port: 443, hostname: configOrNamespaceOrConnectionStringOrSettings + '.' + this.host });
+    var stsendpoint = url.format({ protocol: 'https:', port: 443, hostname: acsNamespace + '.' + this.host });
```
ramya-rao-a: the above change is not required as the host is used in the next few lines in the right way

amarzavery:
Can you bump the patch version of the package in package.json as well? This will make it easier for the team to publish the package.

ramya-rao-a:
@amarzavery The version in the master branch (0.11.0) is already higher than the published version (0.10.6). Do we still have to bump the patch version?

PR #5084, opened 2019-06-07
Update batch data plane SDK to API version 2019-06-01

xingwu1:


adxsdk6:

amarzavery:
Hi @xingwu1 - We are not releasing new versions in node sdk anymore. It has been replaced by the JS sdk which works in node and browser. You can find the batch data plane sdk over here https://github.com/Azure/azure-sdk-for-js/tree/master/sdk/batch/batch.
I ported over the credential changes over there. So the sdk is usable. Please send a PR over there.

PR #5085, opened 2019-06-20
 Fix "mapper.constraints[constraintType].split is not a function" error 

kpajdzik:


adxsdk6:

amarzavery:
Do we have a test for this? If not, then can we add a scenario in the existing set of serialization tests.

kpajdzik:
> Do we have a test for this? If not, then can we add a scenario in the existing set of serialization tests.

There is no single tests for constraints. Can you suggest anything? I don't have any experience with mappers.

PR #5090, opened 2019-06-28
Remove the body initialization for null.

raych1:
Body is not a required property and can be undefined.

### DIFF
--- a/runtime/ms-rest/lib/webResource.js
+++ b/runtime/ms-rest/lib/webResource.js
@@ -25,7 +25,6 @@ class WebResource {
     this.url = null;
     this.method = null;
     this.headers = {};
-    this.body = null;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


adxsdk6:

runtime/ms-rest/lib/webResource.js
```diff
@@ -25,7 +25,6 @@ class WebResource {
     this.url = null;
     this.method = null;
     this.headers = {};
-    this.body = null;
```
amarzavery: Instead of removing it. Please set the body to `undefined`.

amarzavery:
Can you replace the double quote with single quote as mentioned in the CI over [here](https://travis-ci.org/Azure/azure-sdk-for-node/jobs/551604845#L1361).

amarzavery:
Please update the [package.json](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/package.json) version to `2.5.2` and update the [Changelog.md](https://github.com/Azure/azure-sdk-for-node/blob/master/runtime/ms-rest/Changelog.md)

runtime/ms-rest/lib/webResource.js
raych1: Updated.

raych1:
> fix as per comments.

@amarzavery, fix was submitted, pls review.

PR #5097, opened 2019-07-10
do not serialize default values for model properties

amarzavery:


adxsdk6:

Issue #2298, opened 2017-10-13
DDoS vulnerability in dependency (mime)


jornki:
## Recreate
* Create a project and add azure `2.2.1-preview` to it
* Run [nsp](https://github.com/nodesecurity/nsp) against the project
* Security warning: https://nodesecurity.io/advisories/535 

## Result
* Builds depended on nsp checks break

The mime package already has a patch for this vulnerability.

jornki:
Is anything happening with this issue?

PR #5099, opened 2019-07-11
Update out-dated dependencies

### LINKED ISSUE #2457: Broken dependency on node module sax when creating rpm via grunt easy_rpm
This is not exactly azure-sdk's problem, it's just a dependency issue. The same issue is on aws-sdk, https://github.com/aws/aws-sdk-js/issues/435.

using grunt easy_rpm the resulting rpm breaks because of
node_modules/sax/examples on line 1:

#!/usr/local/bin/node-bench

If we can update azure-common's dependency on xml2js to at least 0.2.8, this will solve the problem.

kpajdzik:
Closes #2457 #2298

### DIFF
--- a/lib/common/package.json
+++ b/lib/common/package.json
@@ -24,16 +24,16 @@
   "main": "./lib/common.js",
   "license": "Apache-2.0",
   "dependencies": {
-    "xml2js": "0.2.7",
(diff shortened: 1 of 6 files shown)


adxsdk6:

Issue #4706, opened 2019-02-19
Remove period at beginning of storageEndpointSuffix


EricJizbaMSFT:
My connection strings listed in the portal have this at the end: `EndpointSuffix=core.windows.net`, but `azureEnvironment.js` has an extra period on `storageEndpointSuffix` at the beginning. Using that as EndpointSuffix doesn't work for me.

This doc uses Azure China as an example and seems to back up my suggestion:
https://docs.microsoft.com/azure/storage/common/storage-configure-connection-string

PR #5101, opened 2019-07-11
Remove leading dot from azure environment settings

kpajdzik:


kpajdzik:
Closes https://github.com/Azure/azure-sdk-for-node/issues/4706

adxsdk6:

amarzavery:
@kpajdzik - Can you also fix this in ms-rest-azure-env. It is the same pattern for all the environments. For example: https://github.com/Azure/ms-rest-azure-env/blob/master/lib/azureEnvironment.ts#L280

PR #5113, opened 2019-08-13
Temporarily disable some test case for keyvault because of error: get…

michaeljqzq:
…addrinfo ENOTFOUND sdktestvault74.vault.azure.net


adxsdk6:

PR #5114, opened 2019-08-15
Temporarily remove keyvault test because of error: addrinfo ENOTFOUND…

michaeljqzq:
… sdktestvault74.vault.azure.net


adxsdk6:

PR #5115, opened 2019-08-15
test Batch

michaeljqzq:


adxsdk6:

xingwu1:
@michaeljqzq I updated my PR to fix the code bug in applicationpackageoperations.js which caused by the autorest. You may want to update here to pass the test.

PR #5140, opened 2020-02-20
Fix proxy with credentials

DanielZiv-SF:
tunnel proxy use proxyauth attribute and not auth (that is given by url.prase)
without this fix, setting a proxy to any azure http service (like service bus) with username and password will fail

### DIFF
--- a/lib/common/lib/services/serviceclient.js
+++ b/lib/common/lib/services/serviceclient.js
@@ -864,7 +864,7 @@ ServiceClient.prototype._setDefaultProxy = function () {
       parsedUrl.port = 80;
     }
     if (parsedUrl.auth) {
-      parsedUrl.proxyAuth = parsedUrl.auth
+      parsedUrl.proxyAuth = parsedUrl.auth;
     }
     this.setProxy(parsedUrl);
   } else {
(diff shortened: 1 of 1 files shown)


xirzec:
@jeremymeng You've been working with authenticated proxies recently, can you look at this change? I'm also not sure who actually owns this code at the moment.

jeremymeng:
This change looks fine to me. However I don't know whether we would still use this repository.  It has been deprecated.

ramya-rao-a:
cc @qiaozha

adxsdk6:

DanielZiv-SF:
This package is still in used by our application and is also documented in the official page here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-nodejs-how-to-use-queues
I am aware that there is a different API now (AMQP) but it means we need to change all of the usages in our code..

Do you think you can release this patch as a hotfix to the package?

qiaozha:
> I'm good with it, but don't have merge permissions here.

Merged. Thanks

ramya-rao-a:
@qiaozha I believe @DanielZiv-SF would need us to release an update for the `azure-common` package. Can you do the needful here?

DanielZiv-SF:
@ramya-rao-a @ramya-rao-a Yes please :)

ramya-rao-a:
@qiaozha Any ideas on when the `azure-common` package will be published with the changes in this PR?

qiaozha:
I guess it also needs to bump version and then manually trigger the release pipeline?

PR #5141, opened 2020-03-10
use right class name for import

SvanBoxel:
On line 28 we initiate a client by calling `BillingManagementClient`. However, we import BillingManagement. Update this to be consistent with the call we make.


adxsdk6:

ramya-rao-a:
cc @qiaozha

qiaozha:
ok merged

PR #5143, opened 2020-03-18
ms-rest need to support dash to the parameter name in path template

jianyexi:
Currently, there are specs included parameter name with dash (-) , but ms-rest could not process correctly   the url with segments that contains dash , that  will caused the oav  failed  when swagger reviewing  , because the oav use ms-rest to generate a http-request from an example json file
an example specs:
[KeyVault](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/keyvault/data-plane/Microsoft.KeyVault/stable/7.0/keyvault.json#L82)

### DIFF
--- a/runtime/ms-rest/Changelog.md
+++ b/runtime/ms-rest/Changelog.md
@@ -1,5 +1,11 @@
 # Changelog
+
+## 2.5.4
+
+- Added support for parameter names which contain dash(-) in path template in webResource.
+
 ## 2.5.3
+
 - Do not serialize default values for model properties.
 - During deserialization, set the value of an entity to it's default value if specified in the mapper.
 
--- a/runtime/ms-rest/lib/webResource.js
+++ b/runtime/ms-rest/lib/webResource.js
@@ -169,7 +169,7 @@ class WebResource {
       }
       let baseUrl = options.baseUrl;
       let url = baseUrl + (baseUrl.endsWith('/') ? '' : '/') + (options.pathTemplate.startsWith('/') ? options.pathTemplate.slice(1) : options.pathTemplate);
-      let segments = url.match(/({\w*\s*\w*})/ig);
+      let segments = url.match(/({[\w\-]*\s*[\w\-]*})/ig);
       if (segments && segments.length) {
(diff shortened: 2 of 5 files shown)


adxsdk6:

jianyexi:
> seems fine from my limited understanding, but can you add a test?

I have added it

jianyexi:
@xirzec I have increased the version , updated the changelog,  could you help merge it ?

xirzec:
> @xirzec I have increased the version , updated the changelog, could you help merge it ?

Alas, I do not have the superpowers to merge this. @ramya-rao-a do you have them?

jianyexi:
@ramya-rao-a  could you please take a look ？

ramya-rao-a:
Sorry, I dont have the permissions here either

@qiaozha Can you please help with review, PR merge and eventual publishing of the package?

qiaozha:
Merged. will it be automatically released?

ramya-rao-a:
@qiaozha, I am not aware of what the release process here is.

> that will caused the oav failed when swagger reviewing , because the oav use ms-rest to generate a http-request from an example json file

@jianyexi What is oav? Is the need to support dash in parameter name only for the examples or is this a valid feature request in general?

@xirzec, We should have the same fixed for `@azure/ms-rest-js` and `@azure/core-http` as well if this is a valid feature request in general.

cc @daviwil, @joheredi, @chradek

jianyexi:
@ramya-rao-a  [oav](https://github.com/Azure/oav/tree/master
) is a tool for validating the swagger,  I think  parameter name with dash is valid  as  it has been used in the keyVault service

ramya-rao-a:
We have generated key vault libraries using the same swagger and we haven't seen any issues in running the generated libraries. So, the use in keyvault swagger might be restricted to examples?

jianyexi:
Currently,  only found the impact of model validator in oav tool which will check the examples , but it's better to have a check to the other packages or scripts which refered this package

qiaozha:
The package is still not there. the pipeline I triggered also failed.

PR #5145, opened 2020-03-21
Update Batch data/management plane to new version 2020-03-01

xingwu1:


qiaozha:
Hi, We are on longer release mgmt SDKs in this node repo any more, instead we use https://github.com/Azure/azure-sdk-for-js and the sdk release request should be submit by service team via aka.ms/openapihub

adxsdk6:

PR #5155, opened 2020-05-01
Splitting out Notification Hub library

Petermarcu:
This is a prototype of pulling out the notification hub library from the servicebus library. In all other languages, this was already split out. This will allow people to pull in this notification hubs library separately from servicebus and upgrade them independently when newer ones of one or the other are released. This is intentionally not touching the older servicebus library or trying to take these API's out of it, its explicitly just about having this as a separate option not requiring servicebus as part of getting notification hub.


adxsdk6:

ramya-rao-a:
Update: Pushed commits to fix the below points

Screen Shot 2020-07-21 at 11 16 20 PM

Pending items:
- fix failures in CI: One needs the new package to be published first, but others are unrelated and need to be investigated
- review the code that used values in env vars as backup when none was provided
- run some manual tests against live service.

ramya-rao-a:
Closing this PR in favor of https://github.com/Azure/azure-sdk-for-js/pull/11977
More details in https://github.com/Azure/azure-sdk-for-js/issues/13841

Issue #5153, opened 2020-04-30
Fix the code snippet in the readme for azure-keyvault


jntrnr:
Related to: https://github.com/Azure/azure-sdk-for-node/issues/5151

From Ramya: "the sample for [azure-keyvault](https://github.com/Azure-Samples/key-vault-node-authentication), we see KeyvaultCredentials being used for the data plane library, and the normal login methods of ms-rest-azure for the management library. If the login methods of ms-rest-azure cannot be used directly with azure-keyvault, then the [code snippet in the readme](https://github.com/Azure/azure-sdk-for-node/tree/master/lib/services/keyVault#authentication-and-client-creation) that suggests so should be updated as well. Can you please test this and do the necessary updates?"

sadasant:
I tried to replicate this with my own resources and I haven't been able to do so. So far, azure-sdk-for-node seems to work fine with ms-rest authentication methods. I got in contact with Jon Gallant and he said that he was going to provide me with some specific environment variables that should help me reproduce the bug.

ramya-rao-a:
@sadasant, Can you share which login methods from did you try?

sadasant:
I'll share code snippets as soon as I can go back to this, which should be soon.

sadasant:
Ok so,

What I tried that worked was to use Jon's https://github.com/jongio/storage-node-resource-provider-getting-started

There, Jon uses `KeyVault.msRestAzure.loginWithServicePrincipalSecret` to authenticate.

Since that worked, I tried to reproduce this separately. Here's my isolated code:

```ts
import * as KeyVault from "azure-keyvault";
import * as msRestAzure from "ms-rest-azure";
import { config } from "dotenv";

config();

async function main(): Promise<void> {
  const credentials = await msRestAzure.loginWithServicePrincipalSecret(
    process.env.AZURE_CLIENT_ID,
    process.env.AZURE_CLIENT_SECRET,
    process.env.AZURE_TENANT_ID
  );
  const client = new KeyVault.KeyVaultClient(credentials);
  const keyVaultUrl = `https://${process.env.KEYVAULT_NAME}.vault.azure.net`;
  await client.getKeys(keyVaultUrl);
}

main();
```

That code **fails*. ... But why?

It took me a while, but now I realize that Jon's repo has this definition:

```js
var msRestAzure = require("@azure/ms-rest-nodeauth");
```

Which explains so much 😅

[...]

PR #5167, opened 2020-05-28
[azure-keyvault] Incorrect authentication in README

sadasant:
`azure-keyvault` currently doesn't work with `ms-rest-azure`. It does work with `ms-rest-nodeauth` though.

Here's the code that I'm using to confirm that `ms-rest-azure` isn't working:

```ts
import * as KeyVault from "azure-keyvault";
import * as msRestAzure from "ms-rest-azure";
import { config } from "dotenv";

config();

async function main(): Promise<void> {
  const credentials = await msRestAzure.loginWithServicePrincipalSecret(
    process.env.AZURE_CLIENT_ID,
    process.env.AZURE_CLIENT_SECRET,
    process.env.AZURE_TENANT_ID
  );
  const client = new KeyVault.KeyVaultClient(credentials);
  const keyVaultUrl = `https://${process.env.KEYVAULT_NAME}.vault.azure.net`;
  await client.getKeys(keyVaultUrl);
}

main();
```

Fixes https://github.com/Azure/azure-sdk-for-node/issues/5153

### DIFF
--- a/lib/services/keyVault/README.md
+++ b/lib/services/keyVault/README.md
@@ -34,7 +34,7 @@ For a complete sample, please check one of the above links.
 
 ```javascript
 var KeyVault = require('azure-keyvault');
-var msRestAzure = require('ms-rest-azure');
+var msRestAzure = require('ms-rest-nodeauth');
 
 async function main(): Promise<void> {
   const credentials = await msRestAzure.interactiveLogin();


adxsdk6:

sadasant:
After further inspection, this PR doesn't make sense. For more details, please browse what we've been discussing in this issue: https://github.com/Azure/azure-sdk-for-node/issues/5153#issuecomment-635757198

PR #5171, opened 2020-06-02
[azure-keyvault] Mentioning the ms-rest-azure requirements

sadasant:
It took me a while to figure out what was going on here: https://github.com/Azure/azure-sdk-for-node/issues/5153

From what was concluded in that issue, this PR shows that `azure-keyvault` only supports `ms-rest-azure` version 2.6.0, and it recommends authenticating with `KeyvaultCredentials`, either the one exported by this package (`azure-keyvault`), or the one that comes with `ms-rest-azure` version 2.6.0, while also stating that other authentication methods from `ms-rest-azure` version 2.6.0 are supported.

### DIFF
--- a/lib/services/keyVault/README.md
+++ b/lib/services/keyVault/README.md
@@ -1,8 +1,14 @@
 # Microsoft Azure SDK for Node.js - Key Vault
 
-This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault. Right now it supports:
+This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault.
+
+Right now it supports:
 - **Node.js version: 6.x.x or higher**
 - **REST API version: 7.0**
+- **ms-rest-azure version 2.6.0**.  
+  We recommend authenticating with `KeyvaultCredentials`, either the one exported by this package (`azure-keyvault`),
+  or the one that comes with `ms-rest-azure` version 2.6.0,
+  though other authentication methods from `ms-rest-azure` version 2.6.0 are supported.
 
 ## Features
 
--- a/lib/services/keyVault/README.md
+++ b/lib/services/keyVault/README.md
@@ -2,10 +2,11 @@
 
 This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault.
 
-Right now it supports:
+It supports:
 - **Node.js version: 6.x.x or higher**
-- **REST API version: 7.0**
-- **ms-rest-azure version 2.6.0**
+- **Keyvault REST API version: 7.0**
+
+> Note: Only version 2.6.0 of `ms-rest-azure` package can be used to authenticate. Higher versions of this package are not supported.
 
 ## Features
 
--- a/lib/services/keyVault/README.md
+++ b/lib/services/keyVault/README.md
@@ -5,10 +5,7 @@ This project provides a Node.js package for accessing keys, secrets and certific
 Right now it supports:
 - **Node.js version: 6.x.x or higher**
 - **REST API version: 7.0**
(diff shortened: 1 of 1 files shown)


adxsdk6:

lib/services/keyVault/README.md
```diff
@@ -1,8 +1,14 @@
 # Microsoft Azure SDK for Node.js - Key Vault
 
-This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault. Right now it supports:
+This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault.
+
+Right now it supports:
 - **Node.js version: 6.x.x or higher**
 - **REST API version: 7.0**
+- **ms-rest-azure version 2.6.0**.  
+  We recommend authenticating with `KeyvaultCredentials`, either the one exported by this package (`azure-keyvault`),
```
ramya-rao-a, line 9: What does `KeyvaultCredentials` provide that other login credentials from `azure-keyvault` do not have?

sadasant: I could be understanding what you said wrongly. I am trying to express this what you said here https://github.com/Azure/azure-sdk-for-node/issues/5153#issuecomment-636248372 :

> update readme with a note on when KeyvaultCredentials should be used instead of the credentials from ms rest azure

ramya-rao-a: My bad, I should have been more clear.

From your investigations, what I have understood is that both KeyvaultCredentials and the normal credentials from ms-rest-azure work with azure-keyvault package.

As a user, it is not clear to me as to why or when should I use KeyvaultCredentials vs the other normal credentials from ms-rest-azure. What is the difference between the two? Which of the two are we advocating for?

sadasant: It is as you said, any of them work as long as we're in 2.6.0. I don't personally think we should suggest using one authentication method over the other, though I do think we need to mention the 2.6.0 requirement. I will fix my changes to express what I'm stating in this comment.

sadasant: I mean, I think it would be fair to say only this then?

> Right now it supports:
> - **Node.js version: 6.x.x or higher**
> - **REST API version: 7.0**
> - **ms-rest-azure version 2.6.0**.

ramya-rao-a: I believe it might need a little more attention than clubbing with node version which people usually skip over

lib/services/keyVault/README.md
```diff
@@ -1,8 +1,11 @@
 # Microsoft Azure SDK for Node.js - Key Vault
 
-This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault. Right now it supports:
+This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault.
+
+Right now it supports:
 - **Node.js version: 6.x.x or higher**
 - **REST API version: 7.0**
+- **ms-rest-azure version 2.6.0**
```
ramya-rao-a, line 8: How about this?

```suggestion
This project provides a Node.js package for accessing keys, secrets and certificates on Azure Key Vault.

It supports:
- **Node.js version: 6.x.x or higher**
- **Keyvault REST API version: 7.0**

> Note: Only version 2.6.0 of `ms-rest-azure` package can be used to authenticate. Higher versions of this package are not supported.
```

lib/services/keyVault/README.md
ramya-rao-a: We can handle the note on `KeyvaultCredentials` in a separate PR if needed

lib/services/keyVault/README.md
sadasant: I like it! Thank you!

Issue #5154, opened 2020-05-01
Cannot find module 'moment' when using azure-arm-resource package


caleteeter:
When using the [azure-arm-resource](https://www.npmjs.com/package/azure-arm-resource) package in my typescript project, I am now getting exceptions that the moment module cannot be found.

```
node_modules/azure-arm-resource/lib/feature/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/link/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/lock/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/management/models/index.d.ts:13:25 - error TS2307: Cannot find module 'moment'.

13 import * as moment from 'moment';
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/policy/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/resource/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~

node_modules/azure-arm-resource/lib/subscription/models/index.d.ts:10:25 - error TS2307: Cannot find module 'moment'.

10 import * as moment from "moment";
                           ~~~~~~~~
```

Looking a bit deeper at this, the moment reference is not used in these typings files which are generated from AutoRest Code Generator.  

To test I removed these references to moment in each of the typing files manuallly and the issue is resolved.  Can we have this code generated output fixed to remove packages that are not used?

caleteeter:
Appears to be a break from moment v2.25.0, reverting back to v2.24.0 works fine.

qiaozha:
Thanks for reporting this, I will follow up on this.

qiaozha:
Hi, I notived that the latest version of azure-mgmt-resource is 7.3.0, when I run npm install. the moment package will be include in the package-lock.json Can you give it a try to see if it works?

qiaozha:
[...]

Issue #5157, opened 2020-05-05
ms-rest-azure depends on moment but does not use it


alopix:
The ms-rest-azure tool depends on the moment package but does not actively use it.

[Code Search](https://github.com/Azure/azure-sdk-for-node/search?q=path%3A%2Fruntime%2Fms-rest-azure+moment&unscoped_q=path%3A%2Fruntime%2Fms-rest-azure+moment)

Similarly ms-rest uses it to check if something is a `duration` which could probably achieved using a more lightweight library?
[Code Search](https://github.com/Azure/azure-sdk-for-node/search?q=path%3A%2Fruntime%2Fms-rest+moment&unscoped_q=path%3A%2Fruntime%2Fms-rest+moment)

qiaozha:
Thanks for reporting this, I will follow up on this.

qiaozha:
@ramya-rao-a  Hi Ramya, I found this issue https://github.com/Azure/azure-sdk-for-node/issues/5154 is about the moment package as well. the code referenced moment but actually don't use it. Should we remove the moment dependency and release a new package for azure-arm-resource ?

ramya-rao-a:
For ms-rest-azure, we can definitely remove the dependency altogether

@maggiepint What would you recommend as a replacement for `moment.isDuration()`?

@qiaozha, Issue #5154 is slightly different as I believe that is auto generated code. You can either fix the code generator and re-generate the package or remove the unused moment references manually. Since we don't plan to do any further feature work here and therefore no code regeneration in the future, it should be safe to manually change it.

ramya-rao-a:
@qiaozha, Can we update `ms-rest-azure` to remove the dependency on `moment` which it does not use?

 @maggiepint What would you recommend as a replacement for `moment.isDuration()`?

qiaozha:
> @qiaozha, Can we update `ms-rest-azure` to remove the dependency on `moment` which it does not use?
> 
> @maggiepint What would you recommend as a replacement for `moment.isDuration()`?

I can remove it and I need to release a new ms-rest-azure ?

ramya-rao-a:
I don't believe a release is needed as this is not a bug fix. The next time we have a bug fix, the changes can be released together

qiaozha:
@ramya-rao-a Okay, then I won't release the package unless there's a request specifically ask for a new release. Can you help review my PR #5188

qiaozha:
[...]

PR #5173, opened 2020-06-03
Fix invalid reference

qiaozha:
to fix issue #5154

### DIFF
--- a/lib/services/resourceManagement/package.json
+++ b/lib/services/resourceManagement/package.json
@@ -6,7 +6,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Srinivasan, Vivek <visriniv@microsoft.com>"
   ],
-  "version": "8.0.0",
+  "version": "7.4.0",
   "description": "Microsoft Azure Resource Management Client Library for node",
   "tags": [
     "azure",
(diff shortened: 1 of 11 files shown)


adxsdk6:

lib/services/resourceManagement/package.json
```diff
@@ -6,7 +6,7 @@
     "Zavery, Amar <amzavery@microsoft.com>",
     "Srinivasan, Vivek <visriniv@microsoft.com>"
   ],
-  "version": "7.3.0",
+  "version": "8.0.0",
```
ramya-rao-a, line 9: Why would this change need a major version update?

qiaozha: Just in case some user may use this reference explicitly

qiaozha: Actually I think we should remove the dependency in ms-rest-azure https://github.com/Azure/azure-sdk-for-node/issues/5157 here

ramya-rao-a: Though moment is imported in these files, it is not exported and therefore a user would have no visibility to it. So, a major version update is not needed.

qiaozha: Okay, Changed. Please review it again.

Issue #5181, opened 2020-07-22
v0.9.23 published to NPM an hour ago introduces: Cannot find module 'xmlbuilder/lib/XMLBuilder'


freshlogic:
Error: Cannot find module 'xmlbuilder/lib/XMLBuilder'
Require stack:
- /azure-common/lib/util/patch-xmlbuilder.js

freshlogic:
Reproduction created here: https://github.com/Azure/azure-sdk-for-node/pull/5179#issuecomment-662212263

rnpramasamy:
Yesterday azure-sb was working and Installed this package again in the new package, I got below error

Error: Cannot find module 'xmlbuilder/lib/XMLBuilder'

ChuckLangford:
+1 on this issue

ramya-rao-a:
@qiaozha Can we close this issue now that you have reverted the change that introduced this bug and released an update?

qiaozha:
sure  https://www.npmjs.com/package/azure-common/v/0.9.24

PR #5179, opened 2020-07-16
try fix xml builder

qiaozha:
This is PR is trying to fix issue https://github.com/Azure/azure-sdk-for-node/issues/5159

PR #5182, opened 2020-07-22
fix the xmlbuilder bug

qiaozha:
resolve https://github.com/Azure/azure-sdk-for-node/issues/5181


adxsdk6:

qiaozha:
@ramya-rao-a  @joheredi  Can you help review this PR. Looks like the CI testing fail is because of the bug I am trying to fix.

joheredi:
@qiaozha I'm not familiar with this package, can you please provide some details on what is the bug that this PR is fixing? I see that the error message is `Error: Cannot find module 'xmlbuilder/lib/XMLBuilder'` however I'm not sure I understand what caused it? 

Was this a regression introduced by #5179?

Also it seems like the validation step `continuous-integration/travis-ci/pr` is failing with the same error

qiaozha:
> @qiaozha I'm not familiar with this package, can you please provide some details on what is the bug that this PR is fixing? I see that the error message is `Error: Cannot find module 'xmlbuilder/lib/XMLBuilder'` however I'm not sure I understand what caused it?
> 
> Was this a regression introduced by #5179?
> 
> Also it seems like the validation step `continuous-integration/travis-ci/pr` is failing with the same error

Yes this is a regression introduced by #5179 and the validation step is failing because it's trying to load the latest published azure-common instead of build from the PR's code in the serviceBus dependency. And since I haven't released this fix, the validation step is still using the problematic one. Seems like there's a paradox. 
I have tried to change the serviceBus dependency into my local built azure-common in package.json.
image
 It can pass the test.

ramya-rao-a:
@qiaozha,

Can we consider reverting the changes from #5179, release an update to unblock the customer and then update this PR to do the fix the right way?

ramya-rao-a:
That should help with the CI as well.

qiaozha:
But the msportal is reporting another security bug of the old change

ramya-rao-a:
I am only suggesting the revert temporarily until you figure out exactly what is causing the problem and answer questions that @joheredi has raised.

Teams using msportalfix would at max get the security vulnerability warning for another day or two
But in its current state, azure-common is breaking existing customers

qiaozha:
> I am only suggesting the revert temporarily until you figure out exactly what is causing the problem and answer questions that @joheredi has raised.
> 
> Teams using msportalfix would at max get the security vulnerability warning for another day or two
> But in its current state, azure-common is breaking existing customers

okay, revert the change and re-released https://www.npmjs.com/package/azure-common/v/0.9.24 

The problem is caused by this file https://github.com/Azure/azure-sdk-for-node/pull/5182/files#diff-aae09a2931ddbc9cea60eedb94487627L19 and here https://github.com/Azure/azure-sdk-for-node/pull/5182/files#diff-345af09bd267b08e4253ea69f3330697L21 it has required this file but never used it. So I removed them in this PR.

ramya-rao-a:
> it has required this file but never used it. So I removed them in this PR.

One need not directly use this file for it to be useful.

The file patches the `assertLegalChar()` method in the xml builder.
So, any code in this repo that uses this method or uses some other method in xml builder which then uses this method would have difference in behavior when we remove this patch.

qiaozha:
I also searched assertLegalChar() it was not being used. Just found the CI is still failing not sure why and I can't repro in my local env

ramya-rao-a:
> I also searched assertLegalChar() it was not being used

There is still a possibility that methods from xml builder that we do use may be using assertLegalChar()

qiaozha:
In the 15.1.1 the only place that is using assertLegalChar is in ./node_modules/xmlbuilder/lib/XMLStringifier.js here 
image
But that one has its own definition 
image

qiaozha:
Now all the CI tasks pass.

joheredi:
> In the 15.1.1 the only place that is using assertLegalChar is in ./node_modules/xmlbuilder/lib/XMLStringifier.js here
> image
> But that one has its own definition
> image

@qiaozha as I understand it, the patched `assertLegalChar` removed in this PR was overriding the implementation you pointed out above, so when `xmlbuilder/lib/XMLStringifier.js` uses `assertLegalChar`, it is actually using the patched implementation. Removing the patched implementation would, as @ramya-rao-a mentioned, result in different behavior because now the original `assertLegalChar` function would be used.

I worry about removing the patch as it may cause unexpected behavior for consumers of this library.

The error `Error: Cannot find module 'xmlbuilder/lib/XMLBuilder'` seems to be caused by a refactor of the library. In the new version the function seems to be defined in `xmlbuilder/lib/XMLStringifier`

I think there are 2 ways we can proceed

1) Remove the patch - For this we need to be sure that the current implementation of `assertLegalChar` in XmlBuilder works for us and that the reason the patch was added in the first place has been fixed in XmlBuilder.

2) Update the patched `assertLegalChar` to work with the new XmlBuilder - This will keep the current behavior, a couple changes will be needed (1) replace `'xmlbuilder/lib/XMLBuilder'`  -> `'xmlbuilder/lib/XML'` and return `str` in the patched `assertLegalChar`

I think (2) would be less risky. What do you think?

qiaozha:
@joheredi I do think (2) is a better way. and changed it back.

qiaozha:
I should allso release a new package for legacyStorage as I have changed its way of xmlbuilder.create.

joheredi:
Thanks @qiaozha, I used the following code to repro the issue and try out the changes in the patched function. It may be a good idea to add something similar to the unit tests to make sure we are covered in this front. I'm okay if this test is added in a follow up PR, your call.

```javascript
const common = require("./lib/common");

var x = common.js2xml.createElement("TestElement");

var y = common.js2xml.serializeDocument(x);

console.log(y);

```

qiaozha:
> Thanks @qiaozha, I used the following code to repro the issue and try out the changes in the patched function. It may be a good idea to add something similar to the unit tests to make sure we are covered in this front. I'm okay if this test is added in a follow up PR, your call.
> 
> ```js
> const common = require("./lib/common");
> 
> var x = common.js2xml.createElement("TestElement");
> 
> var y = common.js2xml.serializeDocument(x);
> 
> console.log(y);
> ```

Thanks for point it out, I absolutely can add this.

qiaozha:
> > Thanks @qiaozha, I used the following code to repro the issue and try out the changes in the patched function. It may be a good idea to add something similar to the unit tests to make sure we are covered in this front. I'm okay if this test is added in a follow up PR, your call.
> > ```js
> > const common = require("./lib/common");
> > 
> > var x = common.js2xml.createElement("TestElement");
> > 
> > var y = common.js2xml.serializeDocument(x);
> > 
> > console.log(y);
> > ```
> 
> Thanks for point it out, I absolutely can add this.
Just found there's already test here https://github.com/Azure/azure-sdk-for-node/blob/master/test/util/js2xml-tests.js#L145-L177 so I am going to merge and release the new package next week.

Issue #5176, opened 2020-06-09
[Documentation] Authentication documentation error for azure-graph sdk using interactive authentication


joshlarsen:
### Error Observed

[Documentation/Authentication.md](/Azure/azure-sdk-for-node/blob/master/Documentation/Authentication.md) references `tokenAudience` of `graph`, but it should be `https://graph.microsoft.com`. Using `graph` will result in the error below. Using Service Principal Authentication with the subscription and tenant works fine.

```text
AADSTS500011 The resource principal named 'graph' was not found in the tenant named [currentTenantID].
```

<img width="667" alt="Screenshot 2020-06-09 10 24 58" src="https://user-images.githubusercontent.com/2565382/84193302-396db300-aa69-11ea-86e4-4597c0dd740b.png">

### Recommended Fix

Update the following:

```javascript
const AzureGraphClient = require('azure-graph');
const MsRestAzure = require('ms-rest-azure');

const options = {
  tokenAudience: 'graph',
  domain: '<tenantId>' 
};

MsRestAzure.interactiveLogin(options, (err, credentials) => {
  if (err) throw err;

  let graphClient = AzureGraphClient(credentials, '<tenantId>');

  // ..use the client instance to manage service resources.
});
```

To:

```javascript
const AzureGraphClient = require('azure-graph');
const MsRestAzure = require('ms-rest-azure');

const options = {
  tokenAudience: 'https://graph.microsoft.com',
  domain: '<tenantId>' 
};

MsRestAzure.interactiveLogin(options, (err, credentials) => {
  if (err) throw err;

  let graphClient = AzureGraphClient(credentials, '<tenantId>');

  // ..use the client instance to manage service resources.
});
```

qiaozha:
@ramya-rao-a  Hi Ramya, Is azure-graph a data-plane package?

ramya-rao-a:
@willmtemple Can you take a look at this? Based on your prior work in `@azure/graph`, do you recall what the right value to use here would be?

I am hesitant to make this change without due investigation as the use of `graph` for token audience has been documented in a lot of places and there are hard-coded checks for it as well. Relevant PRs
- #2420
- #1922

joshlarsen:
To clarify, `{ tokenAudience: 'https://graph.microsoft.com' }` is only necessary for interactive authentication. Using `{ tokenAudience: 'graph' }` works for Service Principal authentication (and perhaps other methods).

ramya-rao-a:
[...]

PR #5183, opened 2020-07-23
Update docs to use https://graph.windows.net as tokenAudience for graph

ramya-rao-a:
Based on the findings at https://github.com/Azure/azure-sdk-for-node/issues/5176#issuecomment-658029660, it is not always reliable to use `graph` for `tokenAudience` when working with `azure-graph`.

This PR updates the docs to use `https://graph.windows.net` instead


adxsdk6:

ramya-rao-a:
@willmtemple, Please review.

cc @joshlarsen

ramya-rao-a:
The CI does not like this branch. Created a separate PR in #5187 where all is green.

PR #5185, opened 2020-07-27
update-legacystorage-test-against-the-latest-code

qiaozha:
update the legacy storage test so that they are testing against the latest code instead of publish version.


adxsdk6:

PR #5187, opened 2020-07-27
Update docs to use https://graph.windows.net as tokenAudience for graph

ramya-rao-a:
Creating a PR with same changes as #5183 to test if CI works better here


adxsdk6:

Issue #5208, opened 2021-05-07
Fix for underscore vulnerability


skhilliard:
We are currently using the latest version (0.11.1) of the azure-sb library (https://www.npmjs.com/package/azure-sb).   There is now an underscore vulnerability (via the paths azure-sb > underscore, azure-sb > azure-common > underscore) that needs to be addressed (https://npmjs.com/advisories/1674).

Are there any plans to update this library to patch this?

Thanks,
Kelly

lirenhe:
cc @nickzhums

dw511214992:
@skhilliard please use the latest package. thanks

skhilliard:
@dw511214992 Thank you, I have.  I appreciate everyone's efforts in resolving this issue!

Issue #5207, opened 2021-04-30
azure-sb notificationHubservice.send method using TCP protocol which will causing function app SNAT exhaustion issue when request increase


yuu045-ran:
Hello experts,

**Issue description**

azure-sb notificationHubservice.send method using TCP protocol which will causing function app SNAT exhaustion issue when request count increase, based on network packet capture result, each .send seems to occupy one SNAT port when using function app, this will cause outbound connection issue.(https://devblogs.microsoft.com/premier-developer/reducing-snat-port-consumption-in-azure-app-services/)

**Analysis**

https://github.com/Azure/azure-sdk-for-node/blob/fa35b874c212ed71a56f8093ca8e1d959f661cfb/examples/ASM/notification-hubs.md

Starting to track the .send method from here:
https://github.com/Azure/azure-sdk-for-node/blob/fa35b874c212ed71a56f8093ca8e1d959f661cfb/lib/services/serviceBus/lib/notificationhubservice.js#L136

Eventually, the TCP connection of .send method seems to be by-design because request method is finally used, and I'm not sure if there is a way to force the HTTP client to be re-used or keep alive - 
https://github.com/Azure/azure-sdk-for-node/blob/fa35b874c212ed71a56f8093ca8e1d959f661cfb/lib/common/lib/services/serviceclient.js#L235

**Questions**

- Is the above sympton a known issue for notificationHubservice.send method ? Will we have an ETA on the fix?

- Do we have any known-workarounds to force the TCP connection reuses SNAT port(e.g. similiar to https://www.npmjs.com/package/keep-alive-agent) in code?
- If the issue needs time to be fixed and it's not possible to modify parameters in notificationHubservice.send method, do we have any known workarounds(e.g. using API then reuse API's HTTP client)?

Thank you in advance.

ramya-rao-a:
@mpodwysocki Can you help here?

mpodwysocki:
@yuu045-ran Yes, this is indeed an issue, and although in Node 0.12, connection pooling was added, we are not taking advantage of this.  To enable this, we would need to do the following to use the agent options to use the `http.Agent` class.

```js
request({
    method: method,
    uri: uri,
    agentOptions: {keepAlive:true, maxSockets:1},
    agentClass: http.Agent
})
```

yuu045-ran:
[...]

PR #5209, opened 2021-05-19
Updating Azure-SB with Agent-Options and Keep-Alive behavior

mpodwysocki:
As per issue #5207 we have an issue with SNAT exhaustion as the behavior of the request library we use does not set the keep alive nor the max number of sockets allowed.  This PR allows the two values of `keepAlive` and `maxSockets` to be set through the `ServiceBusServiceClient.prototype.setAgentOptions` method.

This is then used to override the behavior in `ServiceBusServiceClient.prototype._buildRequestOptions` to override the `callback` parameter to add the two properties to the `requestOptions` to modify the behavior of the request going to the `request` module.

### DIFF
--- a/lib/services/serviceBus/lib/servicebusserviceclient.js
+++ b/lib/services/serviceBus/lib/servicebusserviceclient.js
@@ -68,6 +68,8 @@ util.inherits(ServiceBusServiceClient, ServiceClient);
  * @param {object} options The agent options which sets the keepAlive and maxSockets properties.
  */
 ServiceBusServiceClient.prototype.setAgentOptions = function (options) {
+  options || (options = {});
+
   this.agentOptions.keepAlive = options.keepAlive == null ? true : options.keepAlive;
   this.agentOptions.maxSockets = options.maxSockets == null ? Infinity : options.maxSockets;
 };
--- a/lib/services/serviceBus/lib/servicebusserviceclient.js
+++ b/lib/services/serviceBus/lib/servicebusserviceclient.js
@@ -68,7 +68,8 @@ util.inherits(ServiceBusServiceClient, ServiceClient);
  * @param {object} options The agent options which sets the keepAlive and maxSockets properties.
  */
 ServiceBusServiceClient.prototype.setAgentOptions = function (options) {
-  this.agentOptions = options;
+  this.agentOptions.keepAlive = options.keepAlive == null ? true : options.keepAlive;
+  this.agentOptions.maxSockets = options.maxSockets == null ? Infinity : options.maxSockets;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


ramya-rao-a:
Thanks @mpodwysocki!

@jeremymeng, Can you take a look at this PR?

mpodwysocki:
@ramya-rao-a @jeremymeng Looks like we need to fix the CI server since they are using long out of date versions of Node.js as well,

jeremymeng:
> @ramya-rao-a @jeremymeng Looks like we need to fix the CI server since they are using long out of date versions of Node.js as well,

It looks that the typescript version is too low to understand the `/// <reference...` syntax.   Upgrading to v3.5.1 fixes it.

jeremymeng (approved):
LGTM

lib/services/serviceBus/lib/servicebusserviceclient.js
```diff
@@ -50,10 +52,26 @@ function ServiceBusServiceClient(accessKey, issuer, sharedAccessKeyName, sharedA
...
 
 util.inherits(ServiceBusServiceClient, ServiceClient);
 
+/**
+ * Sets the agent options to be sent to the request method which includes the keepAlive and maxSockets.
+ * 
+ * @param {object} options The agent options which sets the keepAlive and maxSockets properties.
+ */
+ServiceBusServiceClient.prototype.setAgentOptions = function (options) {
+  this.agentOptions.keepAlive = options.keepAlive == null ? true : options.keepAlive;
```
jeremymeng, line 71: do we want to validate `options` before accessing its properties?

mpodwysocki: Do you want to check if is boolean for the keepAlive and a number for sockets, I can do that.

jeremymeng: No. I am thinking of checking whether `options` is `undefined` (although it would be rare that customers would call the function without passing in an argument)

mpodwysocki: Ok, how about I fix that with `options || (options = ={});`?

jeremymeng: Yeah something to prevent accessing an undefined would be fine.

mpodwysocki: Updated

jeremymeng (approved):

ramya-rao-a:
@dw511214992,
We merged this PR, but don't have access to releasing the package.
Can you please help with releasing an update to the `azure-sb` package?

dw511214992:
Hi @ramya-rao-a  Sorry for update late as I just get the permission to release it. This PR will be released together with https://github.com/Azure/azure-sdk-for-node/pull/5210, which is used to fix https://github.com/Azure/azure-sdk-for-node/issues/5208. Please review the PR and help merge it. thanks

ramya-rao-a:
Thanks @dw511214992!

Once the update to `azure-sb` is published, please drop a comment in https://github.com/Azure/azure-sdk-for-node/issues/5207 and close that issue

dw511214992:
sure @ramya-rao-a 
Cloud you please help merge the PR #5210 ? We don't have permission to merge it. thanks

PR #5210, opened 2021-05-24
update underscore in azure-sb and azure-common

dw511214992:


dw511214992:
@ramya-rao-a please review it. thanks

lib/services/serviceBus/package.json
```diff
@@ -27,7 +27,7 @@
   "dependencies": {
     "azure-common": "^0.9.19",
     "mpns": "2.1.3",
-    "underscore": "^1.8.3",
```
lirenhe, line 30: Will you also increase the version of serviceBus?

dw511214992: no need, because there is an update which updated it.

lirenhe (approved):

jeremymeng (approved):
Looks good to me. Not sure why CI is failing at Copy to Blob step.

ramya-rao-a (approved):

dw511214992:
Hi @ramya-rao-a  cloud you please merge it? we don't have permission to merge it. thanks

dw511214992:
> Looks good to me. Not sure why CI is failing at Copy to Blob step.

looks like there is something wrong with the authentication of coping to blob

ramya-rao-a:
@dw511214992 Can you please look into the build failure? Am not sure of the consequences of merging the PR with that failure

dw511214992:
> @dw511214992 Can you please look into the build failure? Am not sure of the consequences of merging the PR with that failure

It should caused by the access key of the storage account is not valid. I'm asking who has the permission to access the  storage account.

dw511214992:
@ramya-rao-a it's fixed now. please merge the pr

Issue #5212, opened 2021-11-02
Need ms-rest-azure to reference the new version of adal-node in their dependencies (0.2.2)


WilliamRADFunk:
I'm with the Azure Portal team.

We have a S360 code security bug in relation to a dependency (xmldom) that is being referenced by adal-node (prior to 0.2.x), which in turn is being referenced by ms-rest-azure (latest adal-node version reference is ^0.1.28). The bug wants the version of xmldom to be upgraded from 0.6.0 to 0.7.0, but the problem is that xmldom transitioned from xmldom to @xmldom/xmldom prior to versioning up from 0.6.0. Since adal-node referenced the old way "xmldom" (pre 0.2.x), it's not getting the "@xmldom/xmldom" version. They are now referencing it properly in 0.2.2.

Ultimately, we need the "adal-node": "^0.1.28"  to be changed to something like so "adal-node": "^0.2.2" in ms-rest-azure's package.json dependencies list.

qiaozha:
close this issue as PR merged. @WilliamRADFunk feel free to let me know if you still find issues. Thanks

PR #5213, opened 2021-11-03
resolve-ms-rest-azure-security-issue

qiaozha:
to resolve issue #5212

### DIFF
--- a/.azure-pipelines/client.yml
+++ b/.azure-pipelines/client.yml
@@ -8,7 +8,7 @@ jobs:
   - job: 'Build'
 
     pool:
-      vmImage: 'ubuntu-16.04'
+      vmImage: 'ubuntu-20.04'
 
     steps:
       - task: NodeTool@0
@@ -55,7 +55,7 @@ jobs:
       - 'Build'
 
     pool:
-      vmImage: 'ubuntu-16.04'
+      vmImage: 'ubuntu-20.04'
 
     steps:
       - task: NodeTool@0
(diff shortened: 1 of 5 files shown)


qiaozha:
@ramya-rao-a Could you help review this PR ?

ramya-rao-a (approved):
Changes look good to me

Did you test this with any of the packages to ensure none of the breaking changes in adal-node affect us?

qiaozha:
> Changes look good to me
> 
> Did you test this with any of the packages to ensure none of the breaking changes in adal-node affect us?

I run `npm test` in my local and everything looks good. Is it enough ?

ramya-rao-a:
> I run npm test in my local and everything looks good. Is it enough ?

Please test one of the azure-arm-* packages with the changes being made for ms-rest-azure here

qiaozha:
Just realize all the packages azure-arm-* are using "^2.5.5" in their package,json which means non of them will use this package if we release it. Otherwise we will need to bump and release a new package for all the azure-arm-* packages. @ramya-rao-a  What should we do ?

qiaozha:
By the way, I tested in for storage and it works fine with this new ms-rest-azure.

ramya-rao-a:
> Just realize all the packages azure-arm-* are using "^2.5.5" in their package,json which means non of them will use this package if we release it. Otherwise we will need to bump and release a new package for all the azure-arm-* packages. @ramya-rao-a What should we do ?

Another option is to create a hotfix for ms-rest-azure with version 2.6.1 where you update the adal-node dependency. Given that we have limited investment in the packages in this repo as they are deprecated, this would be a better option in terms of effort.

qiaozha:
Hi @ramya-rao-a , after discussion, in order not to break our customers and make sure this security issue could be resolve at customer side. we will release both 2.6.1 and 3.0.1 for this change, and 2.6.1 will be changed based on 2.6.0 and 3.0.1 will be changed based on 3.0.0.

ramya-rao-a:
That sounds good to me, thanks @qiaozha!

ramya-rao-a:
Can you share the link for the PR for 2.6.1?

qiaozha:
Actually, I already release it, just revert this https://github.com/Azure/azure-sdk-for-node/pull/5101 in my local and bump the adal-node version. I don't think we are going to merge this change in the code ?

ramya-rao-a:
>  don't think we are going to merge this change in the code ?

No, but ideally it would be preferred if a branch was maintained for v2 of ms-rest-azure. What if we had to do another hotfix on top of 2.6.1?

qiaozha:
Okay, I will send out the PR. for the change. Thanks

qiaozha:
@ramya-rao-a here's the PR #5216

PR #5216, opened 2021-11-11
[DoNotMerge] ms-rest-azure-2.6.1 change

qiaozha:
This is the PR to show the change of ms-rest-azure 2.6.1

### DIFF
--- a/runtime/ms-rest-azure/lib/azureEnvironment.js
+++ b/runtime/ms-rest-azure/lib/azureEnvironment.js
@@ -60,7 +60,7 @@ var supportedEnvironments = {
     activeDirectoryGraphResourceId: 'https://graph.windows.net/',
     batchResourceId: 'https://batch.core.windows.net/',
     activeDirectoryGraphApiVersion: '2013-04-05',
-    storageEndpointSuffix: 'core.windows.net',
+    storageEndpointSuffix: '.core.windows.net',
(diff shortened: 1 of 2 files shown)


qiaozha:
@ramya-rao-a Here's the change for 2.6.1, I create this branch from the current main branch, so there'll be no change on the  adal-node version  bump.

ramya-rao-a:
@qiaozha What I meant in https://github.com/Azure/azure-sdk-for-node/pull/5213 was that we should have a branch that reflects the state of things for ms-rest-azure 2.6.1. This will help us make future hotfixes.

- Create a branch `ms-rest-azure-v2` from master from the point where version 2.6.0 of ms-rest-azure was released
- Make the changes required to update adal-node
- In the future use this branch to make v2 hotfixes for ms-rest-azure

qiaozha:
@ramya-rao-a  I get your point, but my concern is that there's a lot of other changes after 3.0.0 commit and I am worried if we revert those changes and leave the branch there, it may cause other potential issues. I think the branch in this PR can correctly reflect what's the code look like in 2.6.1 for ms-rest-azure and at the same time doesn't revert those changes for other packages.  Let me know if that makes sense to you.

ramya-rao-a:
Understood, that works for me

Thanks @qiaozha 

We can close this PR

PR #5220, opened 2021-11-24
fix-security-issue-and-add-credscan

qiaozha:


HarshaNalluru (approved):

HarshaNalluru:
I see they are commented in the `testlist.txt`, we don't intend to run the tests for them? Is it because they are deprecated/moved out of the `-node` repo?

qiaozha:
> I see they are commented in the `testlist.txt`, we don't intend to run the tests for them? Is it because they are deprecated/moved out of the `-node` repo?

I am not sure but from the comment https://github.com/Azure/azure-sdk-for-node/blob/master/test/testlist.txt#L29 in testlist.txt, they are removed because they fail to run those test at that time

PR #5233, opened 2022-04-28
try fix cg issue for ms-rest-azure 3.x.x

qiaozha:
to resolve comments  in #5231

### DIFF
--- a/runtime/ms-rest-azure/package.json
+++ b/runtime/ms-rest-azure/package.json
@@ -26,7 +26,7 @@
   "license": "MIT",
   "dependencies": {
     "adal-node": "^0.2.2",
-    "async": "2.6.4",
+    "async": "^2.6.4",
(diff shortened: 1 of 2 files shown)


runtime/ms-rest-azure/package.json
```diff
@@ -26,7 +26,7 @@
   "license": "MIT",
   "dependencies": {
     "adal-node": "^0.2.2",
-    "async": "2.6.4",
+    "async": "^2.6.4",
```
xirzec, line 29: Can you also update the root package.json as well?

xirzec (approved):
Thanks!

runtime/ms-rest-azure/package.json
qiaozha: updated. Thanks

lirenhe (approved):

